PR #20, opened 2020-10-11
Support search file in windows

liuyangc3:


gaogaotiantian (changes requested):
Hi,

Thank you for the pr, this is a very helpful feature. 

Could you make some changes before I merge it in? 

* ```os.path.isfile``` implies ```os.path.exists```([docs](https://docs.python.org/3/library/os.path.html#os.path.isfile)) so ```match_func``` is not needed here. Just check ```os.path.isfile```
* Can you return ```None``` explicitly so it's easier to understand that you return ```None``` if the file is not found?
* The function name ```search_file_in_path``` is not 100% accurate because you also searched the file in current dir(or abspath). So could you rename it to ```search_file``` or something that does not implying path?
* Instead of returning in the middle of the function on line 25. Could you do ```if "path" in os.environ:``` then search in the ```$PATH```? I think that makes more sense and there's one line we do not need to cover. 

I appreciate a lot for your contribution, if you think the changes above are too much, I can merge it in and change it myself. Just let me know.

liuyangc3:
Thanks for the review I will fix them

PR #144, opened 2021-08-19
Make vizviewer command easy to copy and paste

ogrisel:
Small usability improvement to make it faster / more convenient to open generated report file by copy pasting a single line in one go in the terminal.


gaogaotiantian:
This looks fine to me. A quick tip: if you just want to open the report right after tracing, you can use ``--open`` for ``viztracer``.

PR #146, opened 2021-08-20
Clean up exit & patch code

gaogaotiantian:
Remove flamegraph code

### DIFF
--- a/src/viztracer/main.py
+++ b/src/viztracer/main.py
@@ -389,7 +389,7 @@ class VizUI:
                 if self.options.open:  # pragma: no cover
                     import subprocess
                     subprocess.run(["vizviewer", "--once", os.path.abspath(self.ofile)])
-                exit(0)
+                sys.exit(0)
 
 
 def main():
@@ -397,8 +397,8 @@ def main():
     success, err_msg = ui.parse(sys.argv)
     if not success:
         print(err_msg)
-        exit(1)
+        sys.exit(1)
     success, err_msg = ui.run()
     if not success:
         print(err_msg)
-        exit(1)
+        sys.exit(1)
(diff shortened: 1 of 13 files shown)


src/viztracer/main.py
```diff
@@ -87,11 +86,11 @@ def create_parser(self) -> argparse.ArgumentParser:
...
         parser.add_argument("--log_exception", action="store_true", default=False,
                             help="log all exception when it's raised")
-        parser.add_argument("--log_subprocess", action="store_true", default=False,
+        parser.add_argument("--log_subprocess", action="store_true", default=True,
                             help="log subprocesses")
         parser.add_argument("--subprocess_child", action="store_true", default=False,
                             help=argparse.SUPPRESS)
-        parser.add_argument("--log_multiprocess", action="store_true", default=False,
+        parser.add_argument("--log_multiprocess", action="store_true", default=True,
                             help="log multiprocesses")
```
ogrisel, line 94: This option configuration is weird: since `action="store_true"` and `default=True` it does not seem to be able to disable the multiprocess logging.

Maybe it should be

```python
        parser.add_argument("--ignore-multiprocess", dest="log_multiprocess",
                            action="store_false", default=True,
                            help="Avoid collecting trace data from multiprocess workers")
```

Same remark for `log_subprocess`.

gaogaotiantian: This was on purpose. The reason for this was to make ``--log_multiprocess`` a no-op at this point, so people that are using it or have scripts for it can still use it and have the same effect as before. This is a transition phase and I will remove this option eventually.

I have thought about the ability to disable multi-process logging, and it's hard for me to imagine a scenario where the users will benefit from it, unless of course, the patching code breaks the original function.

I probably will still add the ability to disable the multi-process logging in the future, but for now, I want to see if it works well if the feature is always enabled.

Issue #190, opened 2021-12-06
Add support to release linux aarch64 wheels


odidev:
Problem
---------
On aarch64, `pip install viztracer` builds the wheels from source code and then installs it. It requires the user to have a development environment installed on their system. Also, it takes more time to build the wheels than downloading and extracting the wheels from PyPI.

Resolution
-----------
On aarch64, `pip install viztracer` should download the wheels from PyPI.

@gaogaotiantian, Please let me know your interest in releasing aarch64 wheels. I can help with this.

PR #191, opened 2021-12-06
Add Linux AArch64 wheel build support

odidev:
Added linux aarch64 wheel build support.
Related to https://github.com/gaogaotiantian/viztracer/issues/190, @gaogaotiantian Could you please review this PR?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,13 +6,31 @@ on:
 
 jobs:
   deploy-wheels:
-    name: Deploy wheels on ${{ matrix.os }}
+    name: Deploy wheels on ${{ matrix.os }} for ${{ matrix.arch }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: [37, 38, 39, 310]
         manylinux-image: [manylinux2010, manylinux2014, manylinux_2_24]
+        arch: [auto]
+        include:
+          - os: ubuntu-latest
+            manylinux-image: manylinux2014
+            arch: aarch64
+            python-version: 37
+          - os: ubuntu-latest
+            manylinux-image: manylinux2014
+            arch: aarch64
+            python-version: 38
+          - os: ubuntu-latest
+            manylinux-image: manylinux2014
+            arch: aarch64
+            python-version: 39
+          - os: ubuntu-latest
+            manylinux-image: manylinux2014
+            arch: aarch64
+            python-version: 310
         exclude:
           # manyliunx image is not a valid variation on MacOS and Windows
           - os: macos-latest
@@ -26,6 +44,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Set up QEMU
+        if: ${{ matrix.arch == 'aarch64' }}
+        uses: docker/setup-qemu-action@v1
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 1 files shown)


gaogaotiantian:
Hi,

First of all, thank you for thing about contributing to viztracer, I do really appreciate it.

The problem of aarch64 is, we can't support it without proper testing. That's said, we can't build wheels on platforms that we have not tested on(having wheels on a platform normally implies that we support it). If the user has to build from the source, they know it's probably not well tested and they are on their own.

That's said, I'm not opposed to add aarch64+Linux wheels, but before that, we need tests on the platform. If we can somehow figure out a way to do the tests on aarch64 with github actions, I'm happy to expand the wheels to aarch64.

Please let me know if you have more questions.

Tian

odidev:
I ran tests for aarch64 wheels both locally as well as on GitHub Actions. The tests are passing locally however on GitHub Action observing few failures which looks to be due to emulation. Please have a look at the same: 

**AArch64 wheel testing on GitHub Actions:** https://github.com/odidev/viztracer/actions/runs/1592268770

**AArch64 wheel testing locally:**
- Python3.7: [viztracer_testing_logs_py37.txt](https://github.com/gaogaotiantian/viztracer/files/7735262/viztracer_testing_logs_py37.txt)
- Python3.8: [viztracer_testing_logs_py38.txt](https://github.com/gaogaotiantian/viztracer/files/7735264/viztracer_testing_logs_py38.txt) 
- Python3.9: [viztracer_testing_logs_py39.txt](https://github.com/gaogaotiantian/viztracer/files/7735265/viztracer_testing_logs_py39.txt)
- Python3.10: [viztracer_testing_logs_py310.txt](https://github.com/gaogaotiantian/viztracer/files/7735267/viztracer_testing_logs_py310.txt)

gaogaotiantian:
Okay I see. In that case, as we can not test reliably on aarch64, we can't officially support it. I can understand that you want aarch64 wheels for Linux, if you have the script to build the wheels, I guess I can adopt that and make it clear in README and the docs that we do not support aarch64 officially.

So for the PR, maybe just remove everything in ``python-package.yml``, and leave the stuff in the release script(assuming that part works). I probably need to wait for github actions to support aarch64 before official aarch64 support.

odidev:
CircleCI provides Linux AArch64 build agents. Would you be interested in maintaining another CI as it builds and tests Linux AArch64 wheels without emulation.

gaogaotiantian:
At this point, maintaining an extra CI just for aarch64 seems a bit much. If we have a significant number of users using aarch64 on Linux, we should definitely consider it. This is the first and only issue/pr related to aarch64. If in the future we have more requests on the architecture, I would put some efforts for circleci.

odidev:
It can be useful as -

- The Arm architecture has been available on AWS (Graviton2) for about 2 years now and has seen a lot of adoption by various organizations. The install base of Arm users is growing at a rapid pace.

- Oracle cloud will be releasing Arm based Compute Shapes in the future. Please see [here](https://www.anandtech.com/show/16100/oracle-announces-upcoming-cloud-compute-instances-ice-lake-and-milan-a100-and-altra#:~:text=OCI%E2%80%99s%20Arm%20instances%20will%20clock%20as%20high%20as,%E2%80%93%20OCI%20is%20refreshing%20its%20GPU%20compute%20instances)

- Generally, there is a lot of momentum behind various products switching to Arm based CPUs. For example, the Microsoft surface Pro X is an Arm based device.

gaogaotiantian:
Yeah I understand that there are more than a few ARM64+Linux platforms. However, the reality to me is - how many of the users of VizTracer are using them? So for I got none (except for you of course). Adding CircleCI is not a huge effort by itself, maintaining ARM64 is. There will be architecture specific issues and it would be very difficult for me to maintain it without being able to test it locally. I have Windows/Linux/Mac on x86, but no ARM targets. It would be a very large commitment, as long as I add in the aarch64 for official support.
As for now, the src actually builds on aarch64 according to you, and I can publish the wheels with the emphasis on not officially supporting it. People can figure it out on their own if there's some weird issue.
If in the future, there are a significant number of users on aarc64 and known architecture specific issues, I will add whatever needed for aarch64 + Linux and try to maintain it. Before that, I don't think I am prepared to maintain it as it is too difficult.

odidev:
> As for now, the src actually builds on aarch64 according to you, and I can publish the wheels with the emphasis on not officially supporting it. People can figure it out on their own if there's some weird issue.

Sounds good to me.

Issue #193, opened 2021-12-16
Improove --help of the cli


PeterStolz:
I like viztracer, but the help promp could be greatly improoved.

When running the latest version of viztracer the coommand line output looks like:
```bash
#viztracer -h
usage: viztracer [-h] [--rcfile [RCFILE]]

optional arguments:
  -h, --help         show this help message and exit
  --rcfile [RCFILE]
```   
In my opinion the same help page should be printed when `-h` or `--help` are present, as it is printed when running viztracer without an argument.   
The `--rcfile` option is also described in the extended help page in there so I don't see a reason for this behavior.

gaogaotiantian:
You are right, this is a bug I introduced when I added the pre-parser for rcfile. Your fix seems to be working, but in case of this kind of problem happening in the future, could you add some regression test? You can refer to ``tests/test_cmdline.py``, in ``test_no_file`` I checked ``help`` in the output string. Obviously that was not enough. You can add a ``test_help`` method in that class which confirms that ``viztracer``, ``viztracer -h`` and ``viztracer --help`` all produce useful information. The test cause should not be hard to follow. However, you can always ask me if you have more questions. Fixing a bug is good, but making sure it won't happen again is even better.

However, I'd understand if you don't think that's something you'd want to do. Then I can accept your PR and do the test part myself. Feel free to let me know!

Thank you for your contribution.

PeterStolz:
Hi, I added the testcase in the PR. Let me know if they should be improved.

PR #194, opened 2021-12-16
Fixed help command line options by disabling help in a previous parser

PeterStolz:
Fix for https://github.com/gaogaotiantian/viztracer/issues/193

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -164,8 +164,22 @@ loop.close()
 
 class TestCommandLineBasic(CmdlineTmpl):
     def test_no_file(self):
-        result = self.template(["python", "-m", "viztracer"], expected_output_file=None)
-        self.assertIn("help", result.stdout.decode("utf8"))
+        result = self.template(["python", "-m", "viztracer"], expected_output_file=None).stdout.decode("utf8")
+        self.assertIn("help", result)
+        #  Check for a few more arguments to ensure we hit the intended argumentParser
+        self.assertIn("--output_file", result)
+        self.assertIn("--log_async", result)
+        self.assertIn("--attach", result)
+        self.assertIn("--plugins", result)
+        self.assertIn("--rcfile", result)
+
+    def test_help(self):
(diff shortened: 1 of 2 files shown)


tests/test_cmdline.py
```diff
@@ -164,8 +164,22 @@ async def print_sum(x, y):
 
 class TestCommandLineBasic(CmdlineTmpl):
     def test_no_file(self):
-        result = self.template(["python", "-m", "viztracer"], expected_output_file=None)
-        self.assertIn("help", result.stdout.decode("utf8"))
+        result = self.template(["python", "-m", "viztracer"], expected_output_file=None).stdout.decode("utf8")
```
gaogaotiantian, line 167: ``result`` is used in ``template`` function and in this test file multiple times representing a ``subprocess``. I tend to not use the name for a different kind of object.

I think a better way here is to keep the ``result`` as it was before, and do a 

```
result_stdout = result.stdout.decode("utf-8")
```

You could potentially chain it and skipping the middle variable as well, I just think it would be clearer that way.

tests/test_cmdline.py
```diff
@@ -164,8 +164,22 @@ async def print_sum(x, y):
...
+        #  Check for a few more arguments to ensure we hit the intended argumentParser
+        self.assertIn("--output_file", result)
+        self.assertIn("--log_async", result)
+        self.assertIn("--attach", result)
+        self.assertIn("--plugins", result)
+        self.assertIn("--rcfile", result)
+
+    def test_help(self):
+        """Test that all three options print the same help page"""
+        result = self.template(["python", "-m", "viztracer"], expected_output_file=None).stdout.decode("utf-8")
```
gaogaotiantian, line 178: As I mentioned above, I would rather not to use ``result`` as the variable to store the stdout string in this file. You can either get the stdout data from the result, or use chain and a different variable name than ``result``.

gaogaotiantian (changes requested):
Please refer to the comments I left for the test cases :)

PR #196, opened 2022-01-05
Add trio support to snaptrace.c

richardsheridan:
Trio tasks fit neatly into the framework already set up with `SNAPTRACE_LOG_ASYNC`, only extra imports, attribute checks, and error handling needed.

No tests yet, maybe this weekend?

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,6 @@ jobs:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
         os: [ubuntu-latest, macos-latest, windows-latest]
-        trio: [true, false]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
     steps:
@@ -35,10 +34,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         if (Test-Path -Path '.\requirements-dev.txt' -PathType Leaf) {pip install -r requirements-dev.txt}
-    - name: Install Trio
-      if: matrix.trio
-      run: |
-        pip install trio
     - name: Build dist and test with unittest
       if: matrix.os != 'windows-latest'
       run: |
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,6 +15,7 @@ jobs:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
         os: [ubuntu-latest, macos-latest, windows-latest]
+        trio: [true, false]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
     steps:
@@ -34,6 +35,10 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         if (Test-Path -Path '.\requirements-dev.txt' -PathType Leaf) {pip install -r requirements-dev.txt}
+    - name: Install Trio
+      if: matrix.trio
+      run: |
+        pip install trio
     - name: Build dist and test with unittest
       if: matrix.os != 'windows-latest'
       run: |
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -27,7 +27,7 @@ jobs:
       if: matrix.os != 'windows-latest'
       run: |
         python -m pip install --upgrade pip
-        pip install rich orjson
+        pip install rich orjson trio
         if [ -f requirements-dev.txt ]; then pip install -r requirements-dev.txt; fi
     - name: Install dependencies on Windows
       if: matrix.os == 'windows-latest'
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -335,17 +335,14 @@ snaptrace_tracefunc(PyObject* obj, PyFrameObject* frame, int what, PyObject* arg
                     PyObject* curr_task = Py_None;
                     info->paused = 1;
                     for (int i = 0; i < sizeof(curr_task_getters)/sizeof(curr_task_getters[0]); i++) {
-                        if (curr_task_getters[i] == NULL) {
-                            continue;
-                        }
-                        curr_task = PyObject_CallObject(curr_task_getters[i], NULL);
-                        if (!curr_task) {  // RuntimeError, probably
-                            PyErr_Clear();
-                            curr_task = Py_None;
-                            continue;
-                        }
-                        if (curr_task != Py_None) {
-                            break;
+                        if (curr_task_getters[i] != NULL) {
+                            curr_task = PyObject_CallObject(curr_task_getters[i], NULL);
+                            if (!curr_task) {
+                                PyErr_Clear();  // RuntimeError, probably
+                                curr_task = Py_None;
+                            } else if (curr_task != Py_None) {
+                                break;  // got a valid task
+                            }
                         }
                     }
                     info->paused = 0;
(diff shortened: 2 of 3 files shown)


gaogaotiantian:
Hi,

First of all, thank you for the contribution!

I'm good with most of the code. I will leave some comments on the piece that I think could be restructured a bit.

You are correct that this needs some tests before being merged in. It's no hurry though.

For the test, there are a couple of things that should be considered - 
1. We probably need to test a matrix, i.e. having trio or not vs importing trio or not vs using trio or not.
2. For github actions, the environment set up will be a bit tricky. As of now, you can do things like ``orjson``, which will only be installed on some OS. However, I should probably make this more flexible and get more coverage. You don't have to worry about the flexibility now, for this commit, just make sure trio is installed on some of the OSes.
3. Make sure you ``skipIf`` on the tests using trio.

src/viztracer/modules/snaptrace.c
```diff
@@ -331,6 +334,17 @@ snaptrace_tracefunc(PyObject* obj, PyFrameObject* frame, int what, PyObject* arg
                 if (is_python && is_call && (frame->f_code->co_flags & CO_COROUTINE) != 0) {
                     info->paused = 1;
                     PyObject* curr_task = PyObject_CallObject(asyncio_tasks_current_task, NULL);
+                    if (!curr_task) {  // RuntimeError: no running event loop
```
gaogaotiantian, line 337: I think this piece should be reorganized. We could have future supports for other async libraries besides trio and the code will be messy.

I suggest to do a table based check for ``curr_task``:
```
PyObject* get_curr_task_list[] = {asyncio_tasks_current_task, trio_lowlevel_current_task};
PyObject* curr_task = NULL;
for (int i = 0; i < sizeof()/sizeof(); i++) {
    PyObject* get_curr_task = .. ;
    if (get_curr_task) {
    }
    if (curr_task) {
        break;
    } else {
    }
}
```
This would generate less duplicated code and it would be easier to add extra supports in the future.

You can assign ``curr_task`` to ``Py_None`` if you can't find any. That will stop the trace_func from searching for it in the sub-calls.

gaogaotiantian (changes requested):

src/viztracer/modules/snaptrace.c
richardsheridan: Even though I implemented this I want to play devil's advocate that it might be premature generalization. For instance the obvious next candidate for implementing would be curio, but to get the current task from that backend you have to call an [async flavored function](https://github.com/dabeaz/curio/blob/master/curio/traps.py#L110-L114) so it won't work in this scheme. Also this array doesn't capture whether the task's name is in `task.name` `task.get_name()` or `task.id` or whatever. 🤷 Just thinking out loud.

richardsheridan:
Would you mind approving workflows so I can check the test matrix I just made?

src/viztracer/modules/snaptrace.c
gaogaotiantian: I'm not an expert in Python async, but I don't think it's possible to support an async flavored function in tracefunc. It has to be a fast sync function.

I don't think you need to generalize them to be a specific function. As you can tell, the two functions are almost identical. You can put the code in the for loop, which is short enough that won't create too much confusion, and could save a function call. You can check if ``asyncio_tasks_current_task`` is ``NULL`` even though it's never true. Just to make the code cleaner.

Putting the array as global variable is fine, and I think it should be ``static const``.

.github/workflows/python-package.yml
```diff
@@ -15,6 +15,7 @@ jobs:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
         os: [ubuntu-latest, macos-latest, windows-latest]
+        trio: [true, false]
```
gaogaotiantian, line 18: This is not how I meant by adding trio matrix. This implementation will double the testing time, and if we have more optional dev packages, the testing repeats would grow exponentially.

Ideally, there should be a switch to turn a certain package on/off on specific tests, not all of them. I'll work on that in the future. At mean time, could you remove it from the matrix, and just do trio on Linux/Mac, not Windows, or vice versa, to get some coverage? Like ``orjson``.

tests/test_cmdline.py
```diff
@@ -392,6 +409,25 @@ def check_func(data):
...
 
+    @skipIf(importlib.util.find_spec("trio") is None, reason="Trio-specific test")
+    def test_log_trio(self):
+        def check_func(data):
+            tids = set()
+            for entry in data["traceEvents"]:
+                tids.add(entry["tid"])
+            self.assertGreaterEqual(len(tids), 4)
+
+        if int(platform.python_version_tuple()[1]) >= 7:
```
gaogaotiantian, line 420: The test is good. I realized I used ``platform.python_version_tuple`` in another test(which was written a long time ago). I believe ``sys.version_info`` is preferred, at least by ``mypy``, and clearer. Could you change this line to ``sys.version_info >= (3, 7)``?

gaogaotiantian (changes requested):
Aloha! I'm sorry for the delay. I'm on vacation now(in Hawaii obviously). I don't have much time with the computers every day. This situation would be better in a week or so. Please let me know if you have different opinions for my comments.

src/viztracer/modules/snaptrace.c
richardsheridan: see 1232a19, does it look better?

.github/workflows/python-package.yml
richardsheridan: i set it to run on not-windows since I am testing locally on windows

tests/test_cmdline.py
```diff
@@ -381,7 +398,7 @@ def check_func(data):
                 tids.add(entry["tid"])
             self.assertGreaterEqual(len(tids), 4)
 
-        if int(platform.python_version_tuple()[1]) >= 7:
+        if sys.version_info >= (3, 7):
```
gaogaotiantian, line 401: Thanks for changing this, could you remove ``import platform`` to pass flake8?

src/viztracer/modules/snaptrace.c
```diff
@@ -329,8 +332,22 @@ snaptrace_tracefunc(PyObject* obj, PyFrameObject* frame, int what, PyObject* arg
...
+                        curr_task = PyObject_CallObject(curr_task_getters[i], NULL);
+                        if (!curr_task) {  // RuntimeError, probably
+                            PyErr_Clear();
+                            curr_task = Py_None;
+                            continue;
+                        }
+                        if (curr_task != Py_None) {
+                            break;
+                        }
+                    }
```
gaogaotiantian, line 350: This part is basically what I imagined. Let's clear up some logic.

Line 338, I think for shorter loop body like this, it would be better to write
```
if (curr_task_getters[i] != NULL) {
    // stuff
}
```
than using ``continue``.

line 342, I think ``curr_task`` should either be some useful value, or ``NULL`` with a run-time error? So wouldn't it be better here to have
```
if (!curr_task) {
    PyErr_clear();
    curr_task = Py_None;
    // remove continue here
} else {
    break;
}
```

Actually it might be more logical to have the "correct" case first, do ``break`` in ``if``, and clear errors in ``else``.

gaogaotiantian (changes requested):
Thank you for your efforts and patient! I think we are almost at the point of merging. Just some simple fixes :)

src/viztracer/modules/snaptrace.c
richardsheridan: > line 342, I think `curr_task` should either be some useful value, or `NULL` with a run-time error? So wouldn't it be better here to have

That's not quite right, `asyncio` can return `None` if there is a loop but no task.  I'm not exactly sure when that can happen but we should account for that and not break and try the next task getter.

PR #217, opened 2022-03-19
solve cross-platform command color

mochazi:
Support for windows PowerShell and CMD

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setuptools.setup(
         "Topic :: System :: Logging"
     ],
     python_requires=">=3.6",
-    install_requires=["objprint>=0.1.3"],
+    install_requires=["objprint>=0.1.3","printy>=2.2.0"],
     extras_require={
         "full": ["rich", "orjson"]
     },
--- a/src/viztracer/util.py
+++ b/src/viztracer/util.py
@@ -6,7 +6,7 @@ import errno
 import os
 import re
 from typing import Union
-
+from printy import printy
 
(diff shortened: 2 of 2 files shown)


gaogaotiantian:
Hi, I'll reject the PR(and the one in objprint) for the following reasons:

1. Color is already supported in PowerShell(not 100% on cmd), ANSI color is widely supported and is cross-platform. Next time please confirm that the problem you are solving is actually a problem, and prepare a demo or example to illustrate the issue.
2. You did not "solve" the issue either. You only added the dependency and imported the function, you never used it.
3. VizTracer and objprint do not use external dependency so no other libraries will be required for the features.

I appreciate that you tried to make VizTracer better, but next time please submit an issue first for this kind of PR (which is mentioned in the [contributing](https://github.com/gaogaotiantian/viztracer/blob/master/.github/CONTRIBUTING.md) file).

mochazi:
test1
test2
test3
test4
test5
test6
test7
test8
test9
test10
test11


My windows system PowerShell and CMD can render colors, but the rendering fails when using `objprint`
 Only by activating `printy` can it succeed

gaogaotiantian:
You are correct about the issue. VT in PS seems to be off by default. The solution is not correct though :)

The right way to solve this is to figure out why this works. I think I know why and how to solve it. 

The better way to represent this problem is to submit an issue with the content you just sent in the comment, so that I can fix the problem.

Anyway, I'll work on this and hopefully fix it soon.

gaogaotiantian:
This issue will be fixed by #218 , closing this PR.

PR #241, opened 2022-05-13
Store previous timestamp in per-thread info

colesbury:
When running without the GIL, it's not safe to have concurrent accesses
and modifications to the shared "prev_ts" state.

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -59,7 +59,7 @@ PyObject* trio_lowlevel_module = NULL;
 static PyObject* curr_task_getters[2] = {0};
 
 #if _WIN32
-extern LARGE_INTEGER qpc_freq; 
+LARGE_INTEGER qpc_freq;
 #endif
 
 static struct ThreadInfo* get_thread_info(TracerObject* self)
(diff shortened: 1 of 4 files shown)


colesbury:
@gaogaotiantian This is the first of what I expect to be ~3 PRs to make viztracer thread-safe when running without the GIL. 

Please let me know if this approach is acceptable. The change affects both GIL and no-GIL mode. (Alternatively, I could put the changes behind an `#ifdef Py_NOGIL` flag, but that seemed like it would add more complexity to the code.)

(see https://github.com/colesbury/nogil/issues/58)

gaogaotiantian:
This looks okay. It basically only adds an extra dereference to `info`, as we already have `info` in all `get_ts()` usage anyway. Could you inline `get_system_ts()` function? In that case, maybe move the inlined function to the header file. Alternatively you can move the function to `snaptrace.c`. The former might be slightly better if it is used in other files in the future, but I don't have a strong preference.

colesbury:
Thanks for the quick review. I've moved `get_system_ts()` to `util.h` and market is as `inline`.

gaogaotiantian:
In that case, we might not need the `double get_system_ts(void);` declaration in the header file?

gaogaotiantian:
Ah, and you probably need some extra stuff for `qpc_freq`.

In `util.h`, add:
```
#if _WIN32
#include <windows.h>
extern LARGE_INTEGER qpc_freq;
#endif
```

In `snaptrace.c`, remove `extern` before `LARGE_INTEGER gpc_freq`.

That should make it compile.

gaogaotiantian:
Hey @colesbury , let me know if you need any support from viztracer for 3.13 no-gil build (in case it helps your debugging for multi-thread code).

PR #242, opened 2022-05-16
Make viztracer thread-safe without the GIL

colesbury:
This adds a per-Tracer lock for nogil Python that protects the shared
event node buffer and the metadata linked list.

The event count in the multithread test is also relaxed a bit, due to a 
different number of function calls during thread joining in nogil Python.

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -62,9 +62,12 @@ static PyObject* curr_task_getters[2] = {0};
 LARGE_INTEGER qpc_freq;
 #endif
 
-#ifndef Py_NOGIL
-#define Py_BEGIN_CRITICAL_SECTION(m)
-#define Py_END_CRITICAL_SECTION
+#ifdef Py_NOGIL
+#define SNAPTRACE_THREAD_PROTECT_START(self) Py_BEGIN_CRITICAL_SECTION(&self->mutex)
+#define SNAPTRACE_THREAD_PROTECT_END(self) Py_END_CRITICAL_SECTION
+#else
+#define SNAPTRACE_THREAD_PROTECT_START(self)
+#define SNAPTRACE_THREAD_PROTECT_END(self)
 #endif
 
 static struct ThreadInfo* get_thread_info(TracerObject* self)
@@ -99,7 +102,7 @@ static inline struct EventNode* get_next_node(TracerObject* self)
 {
     struct EventNode* node = NULL;
 
-    Py_BEGIN_CRITICAL_SECTION(&self->mutex);
+    SNAPTRACE_THREAD_PROTECT_START(self);
     node = self->buffer + self->buffer_tail_idx;
     // This is actually faster than modulo
     self->buffer_tail_idx = self->buffer_tail_idx + 1;
@@ -115,7 +118,7 @@ static inline struct EventNode* get_next_node(TracerObject* self)
     } else {
         self->total_entries += 1;
     }
-    Py_END_CRITICAL_SECTION;
+    SNAPTRACE_THREAD_PROTECT_END(self);
 
     return node;
 }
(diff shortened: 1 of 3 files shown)


colesbury:
The lock acquisition in `get_next_node` can be a bit of a bottleneck in nogil Python when tracing a multithreaded program. (It's a no-op in regular CPython). I think this change is still useful, it just means that the relative overhead of viztracer is higher in multithreaded Python programs when running without the GIL.

I experimented with different designs that make EventNode buffer mostly per-thread. Those changes scaled better, but made it difficult to preserve existing aspects of viztracer, such as the precise limits imposed by the "tracer_entries" settings. The changes in this PR are much simpler and preserve the existing design, so I think it's a better starting point.

gaogaotiantian:
Hi, thanks for working on this.

I'm okay with the overall implementation, as it won't affect cpython performance. It's also a simple way for nogil to work. The overhead will be a bit high but we can work on that later.

There are a couple of things that I would change:

1. The current thread-protection implementation uses nogil macro in the code and no-ops them if ``Py_NOGIL`` is not defined. This seems a bit off. I think it would be better to declare a more generic macro to protect multi-threaded code, and use the `nogil` version when `Py_NOGIL` is defined. If, for example, we need to support another `nogil` implementation, we will experience much less hassle. We could use like ``Py_THREAD_PROTECT_START(self)`` and ``Py_THREAD_PROTECT_END(self)``(or some other name that you think is better) in the code(using ``self`` might be more flexible if there are other implementations in the future) and define them to ``nogil`` implementation when ``Py_NOGIL`` is defined. Of course, no-op them in the `#else` case.
2. You protected metadata node in ``snaptrace_dump`` and ``snaptrace_load``. However, the normal node buffer is also used in those functions. In theory, these two functions should only be executed after the trace is stopped. However, to make it more robust, we could consider protect the whole function with the macros. This function should be long so the extra overhead could be ignored. The other threads are not supposed to write to the buffer when the data is being dumped so it does not affect the theoretical performance in real multi-threading code either.
3. I'm okay with the relaxing of the test. 

Let me know if you have any thoughts about this!

colesbury:
I've updated the PR to use `SNAPTRACE_THREAD_PROTECT_START` and `SNAPTRACE_THREAD_PROTECT_END`. I'm using `SNAPTRACE` as the prefix because the `Py` prefix is usually reserved for macros provided by the Python interpreter.

I've also expanded the protected sections in `snaptrace_dump` and `snaptrace_load`.

src/viztracer/modules/snaptrace.c
```diff
@@ -603,6 +613,8 @@ snaptrace_load(TracerObject* self, PyObject* args)
     PyObject* task_dict = NULL;
     PyObject* func_name_dict = PyDict_New();
 
+    SNAPTRACE_THREAD_PROTECT_START(self);
```
gaogaotiantian, line 616: This line should probably be moved to the beginning of the code? As on line 600, ``self->buffer`` is used to initialize ``curr``.

You can move ``curr`` initialization inside the protection as well(to save some time for the unicode string generation?). 

If you prefer declaration before the protection, initialize ``curr`` to ``null`` then assign it inside the protection would also be okay with me.

colesbury: Ooops. I missed that. I'll move it up earlier.

colesbury: I moved it up to immediately after the definition of `lst`. The `lst` declaration needs to be outside because `Py_BEGIN_CRITICAL_SECTION`/`Py_END_CRITICAL_SECTION` currently delimit a C scope and `lst` is returned at the end of the function. (Not clear if delimiting a C scope was a good idea or not...)

From a correctness point of view, it doesn't really matter if most of the definitions and initializations are inside or outside the protection. (Other than the initialization of `curr` as you pointed out).

I also don't think it matters from a performance point of view: I don't expect any contention when calling snaptrace_load/dump. (The only contention I expect is during tracing when calling `get_next_event`.)

gaogaotiantian: Huh, I did not realize that macro will introduce a C scope. Could you comment that under `#ifdef Py_NOGIL`? That is an unobvious thing and could potentially cause some trouble if the person using it is not aware. Just introduce a bit about the mechanism and the things to know when using them.

Other than that, I think the code is good to me.

colesbury: I added a comment before the macro definitions

PR #271, opened 2022-08-29
Fix a bug on the decompressor

xzxcxzxcxzxc:
In the Vcompressor.c, the decompressor part use `fptr = fopen(filename, "r");` to open the binary file. But if a hexadecimal number "**1A**" in the binary file, this will be considered as the end of the document. Once `fread()` found a "1A", it will stop reading and return a zero, which will cause the program to think the file is corrupt. Use `fptr = fopen(filename, "rb");`  will  treat the "1A" as a normal number. It makes the program pass the test.

But when I run `viztracer -o result.json --decompress .\resulead.jsont.cvf`, The program will show
`wrong header 0
wrong header 0
wrong header 0
wrong header 0`. 
I will try to figure out why this is happening later

### DIFF
--- a/src/viztracer/modules/vcompressor/README.md
+++ b/src/viztracer/modules/vcompressor/README.md
@@ -32,7 +32,7 @@ header(header) - pid(pid) - tid(tid) - name(str)
 
 ### FEE
 
-header(header) - pid(pid) - tid(pid) - name(str) - count(uint64) - [start(ts) - dur(ts)]*
+header(header) - pid(pid) - tid(tid) - name(str) - count(uint64) - [start(ts) - dur(ts)]*
 
 ### File info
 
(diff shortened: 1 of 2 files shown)


xzxcxzxcxzxc:
I think maybe I made a repeat commit, but I do not know how to delete the first one. Sorry.

gaogaotiantian:
As I said in the videos, you can not directly work on the branch that you want to merge into (in this case, `trace-log-compressor`). You need to leave that branch clean and branch from it.
```
git checkout trace-log-compressor
git checkout -b trace-log-compressor-somefeature
```

Then you can do a pull request from your `trace-log-compressor-somefeature` branch to my `trace-log-compressor` branch, and you can update your `trace-log-compressor` branch from mine when there are updates.

Each feature will have its own feature branch and you should delete that branch once the feature is merged in.

So, in this case, I can't merge your code in because the commits are messed up. You need a feature branch that only has the code you expect to merge.

xzxcxzxcxzxc:
ok, I will submit a new PR later

PR #272, opened 2022-08-30
Fixed typo

CokeStudios:


gaogaotiantian:
This is indeed a typo. However, `vdb` will be deprecated in `0.16.0` and this piece of documentation will be removed. So as of now we'll just leave it there and won't bother fixing it.

PR #273, opened 2022-08-30
Fixed a bug in the decompressor

xzxcxzxcxzxc:
In the Vcompressor.c, the decompressor part use `fptr = fopen(filename, "r")`; to open the binary file. But if a hexadecimal number "1A" in the binary file, this will be considered as the end of the document. Once `fread()` found a "1A", it will stop reading and return a zero, which will cause the program to think the file is corrupt. Use `fptr = fopen(filename, "rb")`; will treat the "1A" as a normal number. It makes the program pass the test.

But when I run` viztracer -o result.json --decompress .\result.cvf`, The program will show
`wrong header 0 wrong header 0 wrong header 0 wrong header 0`.
I will try to figure out why this is happening later

### DIFF
--- a/src/viztracer/modules/vcompressor/vcompressor.c
+++ b/src/viztracer/modules/vcompressor/vcompressor.c
@@ -212,7 +212,7 @@ vcompressor_decompress(VcompressorObject* self, PyObject* args) {
         return NULL;
     }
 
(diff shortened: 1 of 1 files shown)


xzxcxzxcxzxc:
I create a new branch to commit my changes. Is this correct?

PR #276, opened 2022-08-31
some small changes

JessyTsu1:


JessyTsu1:
The imported modules and packages are formatted using a plugin called **_isort_**, which is based on PEP-8 specification.
Besides that, I found a code snippet that was repeated three times, so I refactored it
_**tempfile.TemporaryDirectory()**_ at the end is where I found a possible try-finally logic problem. I had only ever used _**os.makedirs()**_ before, so I looked through the **_tempfile_** doc and found the perhaps better **_tempfile.TemporaryDirectory()_** method, which will delete itself when it is done.

src/viztracer/event_base.py
```diff
@@ -8,7 +8,7 @@
 
 
 class _EventBase:
-    def __init__(self, tracer: VizTracer, name: str = "", **kwargs):
+    def __init__(self, tracer: Optional[VizTracer], name: str = "", **kwargs):
         self._viztracer_tracer: VizTracer = tracer
```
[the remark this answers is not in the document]
gaogaotiantian: Actually, it is possible for the `tracer` argument to be `None`. This is to make sure code will run normally without VizTracer attached. If you take a look at `test_vizcounter.py`, there are test cases where `trace` is set to `None.

However, @Sefank is correct that we should change more, `_viztracer_tracer` is a good example. So this is not complete.

tests/test_viewer.py
```diff
@@ -319,19 +320,23 @@ def test_browser(self):
         finally:
             os.remove(f.name)
 
+    def basic_judgment(self):
```
gaogaotiantian, line 323: This is indeed a repeating pattern. However, this is more like a common sequential test of a larger sequential test. You can tell that the tests after this common piece is similar but not the same. So abstrating this piece out won't make the test easier to understand(for example, it's not obvious why the test should listen to port 9003).

It is not uncommon for **testcases** to have a common pattern, we don't need to abstract them all.

tests/test_viewer.py
```diff
@@ -373,20 +373,14 @@ def test_directory_flamegraph(self):
...
+            self.basic_judgment()
             time.sleep(2.5)
             resp = urllib.request.urlopen("http://127.0.0.1:9001/old.json")
             self.assertEqual(resp.url, "http://127.0.0.1:9002/")
 
     @unittest.skipIf(sys.platform == "win32", "Can't send Ctrl+C reliably on Windows")
     def test_directory_max_port(self):
-        try:
-            tmp_dir = tempfile.mkdtemp()
+        with tempfile.TemporaryDirectory() as tmp_dir:
```
gaogaotiantian, line 383: I do remember that I explicitly use try .. finally.. here instead of `TemporaryDirectory`. After some second thoughts, this might not be necessary. So this change might be correct.

gaogaotiantian:
Next time please do not name your PR as `some small changes`. Use a more informative name for your changes.

src/viztracer/event_base.py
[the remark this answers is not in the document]
JessyTsu1: Yes, I saw that [the test case](https://github.com/gaogaotiantian/viztracer/blob/master/tests/test_vizcounter.py#L51) at `test_vizcounter.py` `trace` was set to `None`, so I changed them to `Optional`, but not much more for now

tests/test_viewer.py
```diff
@@ -319,19 +320,23 @@ def test_browser(self):
...
+        resp = urllib.request.urlopen("http://127.0.0.1:9001/fib.json")
+        self.assertEqual(resp.url, "http://127.0.0.1:9002/")
+
     @unittest.skipIf(sys.platform == "win32", "Can't send Ctrl+C reliably on Windows")
     def test_directory(self):
         test_data_dir = os.path.join(os.path.dirname(__file__), "data")
         # --use_external_processor won't work with directory
-        self.template(["vizviewer", "--use_external_processor", test_data_dir], success=False, expected_output_file=None)
+        self.template(["vizviewer", "--use_external_processor", test_data_dir], success=False,
+                      expected_output_file=None)
```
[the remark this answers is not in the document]
JessyTsu1: Actually, it is the result of automatic formatting operations in PyCharm. Just for convenience maybe

tests/test_viewer.py
JessyTsu1: I get it

tests/test_viewer.py
JessyTsu1: Yes, I was a bit confused when I saw it. If there is a bug about `tempfile.mkdtemp()`, then there will be no `finish`, but it won't fail I guess, then there is no need to use try...finally...?, so I wanted to put `tmp_dir = tempfile.mkdtemp()` in front of `try` at first.

JessyTsu1:
> Next time please do not name your PR as `some small changes`. Use a more informative name for your changes.

ok...I just have no idea how to name it at the moment, I'll think more next time

Sefank:
> ok...I just have no idea how to name it at the moment, I'll think more next time

Actually, the PR title can be modified by clicking the `Edit` button beside the title on this page, so you can change it THIS time. :D 

As for the commit, I just learned some git actions recently that may also help you. You can use `git commit --amend` to change the commit comment or use `git rebase -i` to interactively split the commit into multiple ones. Since these actions will completely change the commit including its hash, so when you `git push` these commits, you may need to attach a `--force-with-lease` option. Feel free to try!

Issue #269, opened 2022-08-29
Make vizviewer test more stable on Mac on github actions


gaogaotiantian:
Now vizviewer test sometimes fails on nightly validations([example](https://github.com/gaogaotiantian/viztracer/runs/8056885257?check_suite_focus=true)). It's not a real failure, just the port is somehow occupied. As vizviewer does support custom ports, we can maybe check if the port works first, then try to test on it, which will reduce the fake positives in nightly validation.

acehinnnqru:
So can I just try another port?

gaogaotiantian:
Well, not exactly. We should *not* simply try different ports after failing on a port, as there might be some cavets like "it will always fail on the first try". We should use another independent way to confirm a port is working, before we try `vizviewer` on that port.

acehinnnqru:
I see. Could we provide a method to find a free port that is not working and then try to test `Viewer` on it?

gaogaotiantian:
Yes, we can implement a new method in the test class to find a free port first, then test on it.

acehinnnqru:
If we find a free port first, the cases to test `use_external_processor` option can use this port but not 10000. Right?

acehinnnqru:
I melt a problem when I complete my codes: The test_use_external_processor case could not pass. And when I try to execute `vizviewer --use_external_processor temp.json`, it took a long time but nothing happened until I killed it. I don't know what happened.
My device is Macbook Pro M1 Max. I can see the Python took a 100% CPU usage like it melt a dead loop.

gaogaotiantian:
> If we find a free port first, the cases to test `use_external_processor` option can use this port but not 10000. Right?

No, `test_use_external_processor` is another thing. The issue is about the other general tests which usually run on port 9001. You can refer to the example link for a failure case.

acehinnnqru:
> > If we find a free port first, the cases to test `use_external_processor` option can use this port but not 10000. Right?
> 
> No, `test_use_external_processor` is another thing. The issue is about the other general tests which usually run on port 9001. You can refer to the example link for a failure case.

Oh, I see.

gaogaotiantian:
This issue is fixed by #286.

PR #277, opened 2022-09-02
Fix test viewer

acehinnnqru:
This is a pull request to solve issue #269.
Summary:
- Add _find_a_free_port into TestViewer
- Add URL method to Viewer
- Change cases use 9001 port to a free dynamic port found before creating a viewer.

### DIFF
--- a/tests/test_viewer.py
+++ b/tests/test_viewer.py
@@ -144,6 +144,28 @@ class TestViewer(CmdlineTmpl):
         sock.close()
 
         return port
+    
+    @unittest.skipIf(sys.platform == "win32", "Can't send Ctrl+C reliably on Windows")
+    def test_custom_port(self):
+        json_script = '{"file_info": {}, "traceEvents": []}'
+        try:
+            with tempfile.NamedTemporaryFile(mode='w', suffix='.json', delete=False) as f:
+                f.write(json_script)
+            v = Viewer(f.name, port=self._find_a_free_port())
+            try:
+                v.run()
+                time.sleep(0.5)
+                resp = urllib.request.urlopen(v.url())
+                self.assertTrue(resp.code == 200)
+                resp = urllib.request.urlopen(f"{v.url()}/file_info")
+                self.assertEqual(json.loads(resp.read().decode("utf-8")), {})
+                resp = urllib.request.urlopen(f"{v.url()}/localtrace")
+                self.assertEqual(json.loads(resp.read().decode("utf-8")), json.loads(json_script))
+            finally:
+                v.stop()
+        finally:
+            os.remove(f.name)
+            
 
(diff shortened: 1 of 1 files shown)


acehinnnqru:
Hmmm, there may be a problem that where should the _find_a_free_port method put.

In previous commits, I tried to put it in Viewer but I changed it.

acehinnnqru:
Oh, there may exist a problem in mac M1 Max that the `test_use_external_processor` can not pass( I did not change any code or logic).

gaogaotiantian:
Also, it would be great if you run lint first :)

acehinnnqru:
Sorry, I forgot to lint again when I fix an error.

gaogaotiantian:
Interesting. It failed the exact test that I wanted to patch. Forcing me to think about it again actually gives me a new idea.

It's always the same test that failed - `test_once_timeout`. Now I believe the reason was the viewer process could exit(so close the socket) before the test can test on it, because the `--timeout` argument, which will result in a failure. In this case, as we do want to test the timeout feature for the viewer, we should increase the timeout of `test_once_timeout` to make sure it won't exit before the test confirm the status of the socket. Increasing it to 4/5 should do it.

As for the other changes, I think the `url()` piece is good. We should add a new test to test whether viewer could work on a random port, using your new `_find_a_free_port()` method. However, for all the other tests, let's just test on 9001, which is the most common use case for the users.

acehinnnqru:
Yeah, you're right. I did not meet the error after I use `_find_a_free_port` which is weird.

When I saw these cases, I found that the URL or port was hardcoded which is not so good to change when the port or host changed. So, I add a `url` method.

acehinnnqru:
But why I can pass in my own mac?

gaogaotiantian:
It's a racing issue, so the test will *occasionally* fail. If it always fails, I won't check it in to begin with. So, it's normal that it can pass in your local machine.

So, could you change the test fixes where you used `_find_a_free_port()` and add a new test using that just to test a random port? If you want, you can increase the timeout for `test_once_timeout` too.

acehinnnqru:
OK, I will push the changes tonight.

gaogaotiantian (changes requested):

Fred-Black:
At first, I cannot pass test_use_external_processor too, it seems still runs on port9001(I changed 10000 to 9001 and passed).  After I see this request, I tried new code and passed the test, but it seems uses too many ports? for reference:
image

acehinnnqru:
I have completed the changes. You can check them.

gaogaotiantian:
I believe next time you can use "request for review" feature when you finish your changes.

acehinnnqru:
> I believe next time you can use "request for review" feature when you finish your changes.

Sorry, I am new to make PR to a project.

PR #278, opened 2022-09-04
add file info compressor

TTianshun:
add compress file info feature, use zlib to compress source code

### DIFF
--- a/src/viztracer/modules/vcompressor/vc_dump.c
+++ b/src/viztracer/modules/vcompressor/vc_dump.c
@@ -242,6 +242,9 @@ load_events_from_file(FILE* fptr)
                 break;
             case VC_HEADER_FILE_INFO:
                 file_info = load_file_info(fptr);
+                if (!file_info){
+                    goto clean_exit;
+                }
                 PyDict_SetItemString(parsed_events, "file_info", file_info);
                 Py_DECREF(file_info);
                 break;
@@ -266,40 +269,59 @@ clean_exit:
 }
 
 
-int dump_file_info(PyObject* file_info, FILE* fptr){
+int dump_file_info(PyObject* file_info, FILE* fptr)
+{
     const char * buffer = NULL;
     uint64_t content_length = 0;
     uint64_t compression_length = 0;
-    PyObject* json_ret = NULL;
+    PyObject* json_ret = NULL;    
     PyObject* zlib_ret = NULL;
     PyObject* bytes_data = NULL;
-    PyObject* dumps_func = PyObject_GetAttrString(json_module, "dumps");
-    PyObject* json_args = PyTuple_New(1);
-    PyObject* compress_func = PyObject_GetAttrString(zlib_module, "compress");
-    PyObject* zlib_args = PyTuple_New(1);
-
-    if (!dumps_func) {
-        perror("Failed to access json.dumps()");
-        exit(-1);
-    }
-
-    if (!compress_func) {
-        perror("Failed to access zlib.compress()");
-        exit(-1);
-    }
+    PyObject* dumps_func = NULL;
+    PyObject* compress_func = NULL;
+    PyObject* json_args = NULL;
+    PyObject* zlib_args = NULL;
     
+    dumps_func = PyObject_GetAttrString(json_module, "dumps");
+    if (!dumps_func) {
+        goto clean_exit;
+    }
+
+    compress_func = PyObject_GetAttrString(zlib_module, "compress");
+    if (!compress_func) {
+        goto clean_exit;
+    }
+
     // json dumps file_info
+    json_args = PyTuple_New(1);
+    // file_info points to the file_info in json object we compress
+    // PyTuple_SetItem steals the reference, but we don't want to release file_info here
+    // So we need to call Py_INCREF here
     PyTuple_SetItem(json_args, 0, file_info);
     Py_INCREF(file_info); 
     json_ret = PyObject_CallObject(dumps_func, json_args);
+    Py_DECREF(json_args);
+    if (!json_ret){
+        goto clean_exit;
+    }
 
     // convert string to bytes
     bytes_data = PyObject_CallMethod(json_ret, "encode", NULL);
-    content_length = PyBytes_Size(bytes_data);
+    Py_DECREF(json_ret);
+    if (!bytes_data){
+        goto clean_exit;
+    }
 
     // compress bytes data
+    zlib_args = PyTuple_New(1);
+    content_length = PyBytes_Size(bytes_data);
     PyTuple_SetItem(zlib_args, 0, bytes_data);
     zlib_ret = PyObject_CallObject(compress_func, zlib_args);
+    // zlib_args steals bytes_data, so release zlib_args will release bytes_data
+    Py_DECREF(zlib_args);
+    if (!zlib_ret){
+        goto clean_exit;
+    }
     compression_length = PyBytes_Size(zlib_ret);
     buffer = PyBytes_AsString(zlib_ret);
 
@@ -308,19 +330,27 @@ int dump_file_info(PyObject* file_info, FILE* fptr){
     fwrite(&compression_length, sizeof(uint64_t), 1, fptr);
     fwrite(&content_length, sizeof(uint64_t), 1, fptr);
     fwrite(buffer, sizeof(char), compression_length, fptr);
-
+    // release zlib_ret after write, buffer points to zlib_ret->ob_sval;
     Py_DECREF(zlib_ret);
-    Py_DECREF(bytes_data);
-    Py_DECREF(dumps_func);
-    Py_DECREF(json_args);
-    Py_DECREF(compress_func);
-    Py_DECREF(zlib_args);
+
+clean_exit:
+    if (dumps_func){
+        Py_DECREF(dumps_func);
+    }
+    if (compress_func){
+        Py_DECREF(compress_func);
+    }
+
+    if (PyErr_Occurred()) {
+        return -1;
+    }
 
     return 0;
 }
 
 PyObject*
-load_file_info(FILE* fptr){
+load_file_info(FILE* fptr)
+{
     unsigned char* compression_buffer = NULL;
     uint64_t compression_length = 0;
     uint64_t content_length = 0;
@@ -328,51 +358,61 @@ load_file_info(FILE* fptr){
     PyObject* zlib_ret = NULL;
     PyObject* bytes_data = NULL;
     PyObject* string_data = NULL;
-    PyObject* loads_func = PyObject_GetAttrString(json_module, "loads");
-    PyObject* json_args = PyTuple_New(1);
-    PyObject* decompress_func = PyObject_GetAttrString(zlib_module, "decompress");
-    PyObject* zlib_args = PyTuple_New(1);
-
-    if (!loads_func) {
-        perror("Failed to access json.loads()");
-        exit(-1);
-    }
-
-    if (!decompress_func) {
-        perror("Failed to access zlib.decompress()");
-        exit(-1);
-    }
-
+    PyObject* loads_func = NULL;
+    PyObject* json_args = NULL;
+    PyObject* decompress_func = NULL;
+    PyObject* zlib_args = NULL;
     READ_DATA(&compression_length, uint64_t, fptr);
     READ_DATA(&content_length, uint64_t, fptr);
 
+    decompress_func = PyObject_GetAttrString(zlib_module, "decompress");
+    if (!decompress_func){
+        goto clean_exit;
+    }
+
+    loads_func = PyObject_GetAttrString(json_module, "loads");
+    if (!loads_func){
+        goto clean_exit;
+    }
+
     // read compressed data
     compression_buffer = (unsigned char*)malloc(sizeof(char) * compression_length);
-    if(compression_buffer == NULL){
+    if (compression_buffer == NULL){
+        PyErr_Format(PyExc_RuntimeError, "Failed to malloc memory size %lld", compression_length);
         goto clean_exit;
     }
     fread(compression_buffer, sizeof(char), compression_length, fptr);
     bytes_data = PyBytes_FromStringAndSize((const char *)compression_buffer, compression_length);
-
-    // decompress data
-    PyTuple_SetItem(zlib_args, 0, bytes_data);
-    zlib_ret = PyObject_CallObject(decompress_func, zlib_args);
-
-    // convert depressed bytes to json
-    string_data = PyObject_CallMethod(zlib_ret, "decode", NULL);
-    PyTuple_SetItem(json_args, 0, string_data);
-    file_info = PyObject_CallObject(loads_func, json_args);
-    Py_DECREF(zlib_ret);
     free(compression_buffer);
 
-clean_exit:
-    Py_DECREF(loads_func);
+    // decompress data
+    zlib_args = PyTuple_New(1);
+    PyTuple_SetItem(zlib_args, 0, bytes_data);
+    zlib_ret = PyObject_CallObject(decompress_func, zlib_args);
     Py_DECREF(zlib_args);
+
+    // decode depressed bytes to string
+    string_data = PyObject_CallMethod(zlib_ret, "decode", NULL);
+    Py_DECREF(zlib_ret);
+    if (!string_data){
+        goto clean_exit;
+    }
+
+    // convert string to json
+    json_args = PyTuple_New(1);
+    PyTuple_SetItem(json_args, 0, string_data);
+    file_info = PyObject_CallObject(loads_func, json_args);
     Py_DECREF(json_args);
-    Py_DECREF(decompress_func);
+
+clean_exit:
+    if (loads_func){
+        Py_DECREF(loads_func);
+    }
+    if (decompress_func){
+        Py_DECREF(decompress_func);
+    }
 
     if (PyErr_Occurred()) {
-        Py_DECREF(file_info);
         return NULL;
     }
 
--- a/src/viztracer/modules/vcompressor/vc_dump.c
+++ b/src/viztracer/modules/vcompressor/vc_dump.c
@@ -311,6 +311,13 @@ int dump_file_info(PyObject* file_info, FILE* fptr)
     if (!bytes_data){
         goto clean_exit;
     }
+    // Make sure that PyBytes_Size succeed
+    if (!PyBytes_Check(bytes_data)){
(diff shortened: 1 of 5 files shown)


gaogaotiantian:
It seems like you just rearranged the data format. If there is no repeat information for `file_info`, why not just convert the object to json string and compress it? Having a separate entry for function and file seems an overkill to me. Especially considering in a practical case, when the file really needs compression(> 100MB), `file_info` might be just a fraction. We do not need to put too much effort into compressing it with a perfect compressing issue. It's more important to make the implementation correct, easy to understand and maintainable.

TTianshun:
I've updated file info format, like you said, convert the object to json string and compress it. And it seems better on file size.
the older file info format is like some package format, which is not appropriate for this.
By the way, I'm not sure if it's necessary to add hash verify for uncompressed content.

TTianshun:
> It seems like you just rearranged the data format. If there is no repeat information for `file_info`, why not just convert the object to json string and compress it? Having a separate entry for function and file seems an overkill to me. Especially considering in a practical case, when the file really needs compression(> 100MB), `file_info` might be just a fraction. We do not need to put too much effort into compressing it with a perfect compressing issue. It's more important to make the implementation correct, easy to understand and maintainable.

I've updated file info format, like you said, convert the object to json string and compress it. And it seems better on file size.
the older file info format is like some package format, which is not appropriate for this.
By the way, I'm not sure if it's necessary to add hash verify for uncompressed content.

gaogaotiantian:
Before I start further reviewing, I suddenly have a question - will `zlib` always be available on all system? Are there any prerequisites for using it? I would actually prefer using something from Python if that's possible.

TTianshun:
> Before I start further reviewing, I suddenly have a question - will `zlib` always be available on all system? Are there any prerequisites for using it? I would actually prefer using something from Python if that's possible.

`zlib` is one of the external libs when building python from source, also python has a zlib module, so it should be available available on most system.

gaogaotiantian:
So obviously the CI failed, but I can't figure out why quickly. It seems like something is definitely wrong on Linux and the build process failed on Mac. I'm still not sure if zlib is the solution here. Maybe a Python wrapped library would be less painful.

TTianshun:
> So obviously the CI failed, but I can't figure out why quickly. It seems like something is definitely wrong on Linux and the build process failed on Mac. I'm still not sure if zlib is the solution here. Maybe a Python wrapped library would be less painful.

It seems uLongf is different in some platforms. I committed some fix but I don't have a mac to verify.

TTianshun:
I'll try to use zlib module to compress the file_info. But the fix above changes less for now.

src/viztracer/modules/vcompressor/vcompressor.c
```diff
@@ -220,6 +229,14 @@ vcompressor_decompress(VcompressorObject* self, PyObject* args) {
 
     parsed_events = load_events_from_file(fptr);
 
+    file_info = load_file_info(fptr);
```
gaogaotiantian, line 232: I know `load_events_from_file` might be a bit confusing and ambiguous. Here `events` does not mean `traceEvents`. It simply means "the data from the actual json report"(compared to say metadata which is not part of the json report). So `file_info` should also be read in `load_events_from_file`, as `file_info` should be considered as an "event" in the compressed binary file.

gaogaotiantian (changes requested):

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +261,123 @@ load_events_from_file(FILE* fptr)
...
+    PyObject* dumps_func = PyObject_GetAttrString(json_module, "dumps");
+    PyObject* json_args = PyTuple_New(1);
+    PyObject* zlib_module_name = PyUnicode_FromString("zlib");
+    PyObject* zlib_module = PyImport_Import(zlib_module_name);
+    PyObject* compress_func = PyObject_GetAttrString(zlib_module, "compress");
+    PyObject* zlib_args = PyTuple_New(1);
+    
+    // json dumps file_info
+    PyTuple_SetItem(json_args, 0, file_info);
+    Py_INCREF(file_info);
```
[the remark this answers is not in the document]
TTianshun: Thanks a lot for reviewing, but the file_info here points the file_info in json object that we compress. If `Py_INCREF(file_info)` is removed, the file_info in json object will be released after compress because `json_args` is released.

TTianshun:
I’ve made some changes according to your review.

TTianshun:
I’ve made some changes according to your review. I’m not sure if my changes are right.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+    PyObject* json_ret = NULL;
+    PyObject* zlib_ret = NULL;
+    PyObject* bytes_data = NULL;
+    PyObject* dumps_func = PyObject_GetAttrString(json_module, "dumps");
+    PyObject* json_args = PyTuple_New(1);
+    PyObject* compress_func = PyObject_GetAttrString(zlib_module, "compress");
+    PyObject* zlib_args = PyTuple_New(1);
+
+    if (!dumps_func) {
+        perror("Failed to access json.dumps()");
```
gaogaotiantian, line 282: Do not do `perror` then `exit`. This will confuse Python users as their program will just exit. It's hard to debug and it's impossible to catch such errors in Python. Check the error handling from other code. Do `PyErr_SetString` and `goto` a `cleanup` label(where resources are released in case of an error).

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+    }
+
+    if (!compress_func) {
+        perror("Failed to access zlib.compress()");
+        exit(-1);
+    }
+    
+    // json dumps file_info
+    PyTuple_SetItem(json_args, 0, file_info);
+    Py_INCREF(file_info);
```
gaogaotiantian, line 293: Could you add a comment here indicating that `PyTuple_SetItem` steals reference so we need to do `Py_INCREF`?

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+
+    if (!compress_func) {
+        perror("Failed to access zlib.compress()");
+        exit(-1);
+    }
+    
+    // json dumps file_info
+    PyTuple_SetItem(json_args, 0, file_info);
+    Py_INCREF(file_info); 
+    json_ret = PyObject_CallObject(dumps_func, json_args);
```
gaogaotiantian, line 294: This could fail so it needs error checking.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+
+    // convert string to bytes
+    bytes_data = PyObject_CallMethod(json_ret, "encode", NULL);
```
gaogaotiantian, line 297: This could fail too.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+    content_length = PyBytes_Size(bytes_data);
+
+    // compress bytes data
+    PyTuple_SetItem(zlib_args, 0, bytes_data);
+    zlib_ret = PyObject_CallObject(compress_func, zlib_args);
```
gaogaotiantian, line 302: This could fail.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+    compression_length = PyBytes_Size(zlib_ret);
+    buffer = PyBytes_AsString(zlib_ret);
+
+    // write data
+    fputc(VC_HEADER_FILE_INFO, fptr);
+    fwrite(&compression_length, sizeof(uint64_t), 1, fptr);
+    fwrite(&content_length, sizeof(uint64_t), 1, fptr);
+    fwrite(buffer, sizeof(char), compression_length, fptr);
+
+    Py_DECREF(zlib_ret);
```
gaogaotiantian, line 312: It's better to release the resource as soon as it's no longer in use. For example, `bytes_data`'s referece is stolen by `PyTuple_SetItem(zlib_args, 0, bytes_data)`, so releasing `zlib_args` should release `bytes_data` as well. Releasing everyting at end is making it very difficult to verify the logic.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+    Py_DECREF(bytes_data);
+    Py_DECREF(dumps_func);
+    Py_DECREF(json_args);
+    Py_DECREF(compress_func);
+    Py_DECREF(zlib_args);
+
+    return 0;
```
gaogaotiantian, line 319: `dump_file_info` should return a non-zero value if it failed, and it should be checked.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+    PyObject* file_info = NULL;
+    PyObject* zlib_ret = NULL;
+    PyObject* bytes_data = NULL;
+    PyObject* string_data = NULL;
+    PyObject* loads_func = PyObject_GetAttrString(json_module, "loads");
+    PyObject* json_args = PyTuple_New(1);
+    PyObject* decompress_func = PyObject_GetAttrString(zlib_module, "decompress");
+    PyObject* zlib_args = PyTuple_New(1);
+
+    if (!loads_func) {
```
gaogaotiantian, line 336: Same here, do not use perror and exit.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +264,117 @@ load_events_from_file(FILE* fptr)
...
+    compression_buffer = (unsigned char*)malloc(sizeof(char) * compression_length);
+    if(compression_buffer == NULL){
+        goto clean_exit;
+    }
+    fread(compression_buffer, sizeof(char), compression_length, fptr);
+    bytes_data = PyBytes_FromStringAndSize((const char *)compression_buffer, compression_length);
+
+    // decompress data
+    PyTuple_SetItem(zlib_args, 0, bytes_data);
+    zlib_ret = PyObject_CallObject(decompress_func, zlib_args);
```
gaogaotiantian, line 359: Check if this succeed and handle error accordingly.

src/viztracer/modules/vcompressor/vcompressor.c
```diff
@@ -184,6 +188,11 @@ static PyObject* vcompressor_compress(VcompressorObject* self, PyObject* args)
 
     dump_parsed_trace_events(parsed_events, fptr);
 
+    file_info = PyDict_GetItemString(raw_data, "file_info");
+    if(file_info != NULL){
```
gaogaotiantian, line 192: Add a space after `if`.

src/viztracer/modules/vcompressor/vcompressor.c
```diff
@@ -184,6 +188,11 @@ static PyObject* vcompressor_compress(VcompressorObject* self, PyObject* args)
 
     dump_parsed_trace_events(parsed_events, fptr);
 
+    file_info = PyDict_GetItemString(raw_data, "file_info");
```
gaogaotiantian, line 191: `file_info` is never released right?

gaogaotiantian (changes requested):
Probably should take a deeper look at the error handling and resource release part. It's not easy to get it right and it is one of the most painful part when you do c-python extension. However, that's the price to make this fast.

TTianshun:
> Probably should take a deeper look at the error handling and resource release part. It's not easy to get it right and it is one of the most painful part when you do c-python extension. However, that's the price to make this fast.

I made some changes:

TTianshun:
> Probably should take a deeper look at the error handling and resource release part. It's not easy to get it right and it is one of the most painful part when you do c-python extension. However, that's the price to make this fast.

Hello, thanks for reviewing. I made some changes:
1. For `PyObject` that is used only once, I initialize it when it's used and release it immediately when it's not in use.
2. For `PyObject` that might be used many times, such as `Method`, I initialize it at the beginning of the function and release it at clean_exit.
3. I didn't use `PyErr_SetString` because I found there's error check in `PyObject_GetAttr`, when I use `PyObject_CallObject` incorrectly, there's still error message. I think we shouldn't cover the original error message.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +267,160 @@ load_events_from_file(FILE* fptr)
...
+    PyTuple_SetItem(json_args, 0, file_info);
+    Py_INCREF(file_info); 
+    json_ret = PyObject_CallObject(dumps_func, json_args);
+    Py_DECREF(json_args);
+    if (!json_ret){
+        goto clean_exit;
+    }
+
+    // convert string to bytes
+    bytes_data = PyObject_CallMethod(json_ret, "encode", NULL);
```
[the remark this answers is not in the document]
gaogaotiantian: Also, I know it's a long shot, but we should probably also check `bytes_data` is a `PyByte`. We assumed `PyBytes_Size` will succeed and that relies on the fact that `bytes_data` is a `PyByte`. I know `encode` should always return a byte on success, but this could cause a segfault when say in the future, encode somehow returns something else.

Let's do a `if (!PyBytes_Check(bytes_data))` for error handling just to be safe. I'm okay if my code raises exceptions on something I do not fully cover, but segfault is the thing we should try to avoid.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +267,160 @@ load_events_from_file(FILE* fptr)
...
+    Py_DECREF(json_ret);
+    if (!bytes_data){
+        goto clean_exit;
+    }
+
+    // compress bytes data
+    zlib_args = PyTuple_New(1);
+    content_length = PyBytes_Size(bytes_data);
+    PyTuple_SetItem(zlib_args, 0, bytes_data);
+    zlib_ret = PyObject_CallObject(compress_func, zlib_args);
```
gaogaotiantian, line 319: Same here, check `zlib_ret` is a `byte` after `NULL` checking.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +267,160 @@ load_events_from_file(FILE* fptr)
...
+    }
+
+    // read compressed data
+    compression_buffer = (unsigned char*)malloc(sizeof(char) * compression_length);
+    if (compression_buffer == NULL){
+        PyErr_Format(PyExc_RuntimeError, "Failed to malloc memory size %lld", compression_length);
+        goto clean_exit;
+    }
+    fread(compression_buffer, sizeof(char), compression_length, fptr);
+    bytes_data = PyBytes_FromStringAndSize((const char *)compression_buffer, compression_length);
```
gaogaotiantian, line 385: Check if this succeed as well. `PyBytes_FromStringAndSize` can fail and return a `NULL` pointer.

gaogaotiantian: And of course it's a byte.

gaogaotiantian: And it's never released.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -258,3 +267,160 @@ load_events_from_file(FILE* fptr)
...
+    zlib_args = PyTuple_New(1);
+    PyTuple_SetItem(zlib_args, 0, bytes_data);
+    zlib_ret = PyObject_CallObject(decompress_func, zlib_args);
+    Py_DECREF(zlib_args);
+    if(!zlib_ret){
+        goto clean_exit;
+    }
+
+    // decode depressed bytes to string
+    string_data = PyObject_CallMethod(zlib_ret, "decode", NULL);
```
gaogaotiantian, line 398: This is never released either.

src/viztracer/modules/vcompressor/vcompressor.c
```diff
@@ -184,6 +188,14 @@ static PyObject* vcompressor_compress(VcompressorObject* self, PyObject* args)
 
     dump_parsed_trace_events(parsed_events, fptr);
 
+    // file_info here should not be released
```
gaogaotiantian, line 191: Okay this is a borrowed reference. Could you add this (more explicitly than should not be released) to the comment? Just say this is a borrowed reference would be fine.

gaogaotiantian (changes requested):

src/viztracer/modules/vcompressor/vc_dump.c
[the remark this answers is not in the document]
TTianshun: > Seems like `bytes_data` is never released.

I didn't call `Py_DECREF(bytes_data);` because `zlib_args` steals the reference of `bytes_data`. `Py_DECREF(zlib_args);` will release it.

src/viztracer/modules/vcompressor/vc_dump.c
TTianshun: `bytes_data` is stolen by `zlib_args` here. 
For `PyBytes_FromStringAndSize`, I'll add a check.

src/viztracer/modules/vcompressor/vc_dump.c
TTianshun: `string_data` is stolen by `json_args`

src/viztracer/modules/vcompressor/vc_dump.c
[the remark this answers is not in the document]
gaogaotiantian: > > Seems like `bytes_data` is never released.
> 
> I didn't call `Py_DECREF(bytes_data);` because `zlib_args` steals the reference of `bytes_data`. `Py_DECREF(zlib_args);` will release it.

Right. Tuple is tricky.

TTianshun:
Added `PyBytes_Check` before `PyBytes_Size`.
I also added content length check to make sure the file is not corrupted.

gaogaotiantian:
Thank you for the hardwork on this issue, this is not an easy one. Contrats to your contribution!

PR #268, opened 2022-08-28
VCompressor Benchmark

Sefank:
As #249 mentioned, VCompressor is designed for a decent compress ratio. Here provides a benchmark framework to easily calculate performance improvement space-wise and time-wise on specific test cases.

Note: Here, I chose to use LZMA, famous for its high compress ratio on average files, as a baseline in terms of space-wise benchmark.

---

**Main Features**:
- [x] Space-wise Benchmark
- [x] Time-wise Benchmark
- [x] Display compress ratio in the vcompressor benchmark
- [x] Support multiple other compressors for comparison

**Other Important Change Log**:
- [x] FIX: custom `repeat`(previously `loop_time`) in `_benchmark` is now implemented correctly
- [x] STYLE: All related functions and types are now placed in a single class, without affecting global namespace

---
**Output Example of this Benchmark Test**:
```
2022-09-05 23:01:24,684 INFO: tests.test_vcompressor.TestVCompressorPerformance.test_benchmark_basic start
2022-09-05 23:01:26,302 INFO: 1. On file "vdb_basic.json":
2022-09-05 23:01:26,302 INFO:     [Space]
2022-09-05 23:01:26,302 INFO:       Uncompressed:       2.95KB
2022-09-05 23:01:26,302 INFO:       VCompressor:      531.00B(1.000) [CR: 17.57%]
2022-09-05 23:01:26,302 INFO:       LZMA:             504.00B(0.949) [CR: 16.68%]
2022-09-05 23:01:26,303 INFO:     [Time]
2022-09-05 23:01:26,303 INFO:       VCompressor:        0.405s(1.000)
2022-09-05 23:01:26,303 INFO:       LZMA:               0.004s(0.010)
2022-09-05 23:01:27,757 INFO: 2. On file "multithread.json":
2022-09-05 23:01:27,757 INFO:     [Space]
2022-09-05 23:01:27,757 INFO:       Uncompressed:     105.22KB
2022-09-05 23:01:27,757 INFO:       VCompressor:        9.05KB(1.000) [CR:  8.60%]
2022-09-05 23:01:27,757 INFO:       LZMA:              16.66KB(1.840) [CR: 15.83%]
2022-09-05 23:01:27,757 INFO:     [Time]
2022-09-05 23:01:27,757 INFO:       VCompressor:        0.330s(1.000)
2022-09-05 23:01:27,757 INFO:       LZMA:               0.023s(0.071)
[...]

PR #280, opened 2022-09-09
FIX: Prevent generating `cmdline_test.py` in vcompressor benchmark

Sefank:
I've forgotten cleanup code in #268 so that after benchmark is conducted, there is a `cmdline_test.py` left. Sorry for that :D
This commit should fix it.

### DIFF
--- a/tests/test_vcompressor.py
+++ b/tests/test_vcompressor.py
@@ -132,7 +132,7 @@ class TestVCompressorPerformance(CmdlineTmpl):
     def _benchmark_vcompressor(self, uncompressed_file_path: str, compressed_file_path: str) -> None:
         self.template(
             ["viztracer", "-o", compressed_file_path, "--compress", uncompressed_file_path],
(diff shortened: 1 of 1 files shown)


tests/test_vcompressor.py
```diff
@@ -156,3 +156,4 @@ def test_benchmark_basic(self):
                 vcompress_result = self._benchmark_vcompressor(path)
                 self._print_result(filename, original_size,
                                    vcompress_result, other_results, subtest_idx=subtest_idx)
+        self.cleanup()
```
gaogaotiantian, line 159: The `cmdline_test.py` script is created from line 133, and it is never used. I believe if you just pass `script=None` to `self.template()`, the file won't be created to begin with.

gaogaotiantian (changes requested):

tests/test_vcompressor.py
Sefank: OK. I'll correct it.

PR #281, opened 2022-09-09
Add coverage tests for vcompressor

youker-Shawn:
I try to add some corner test cases for vcompressor. Now the two functions `run_compress` and `run_decompress` are tested with 100% coverage.


tests/data/fib
```diff
@@ -0,0 +1 @@
+{"traceEvents": [{"pid": 2783, "tid": 2783, "ts": 6556500799.741, "dur": 0.3, "name": "<module> (/home/gaogaotiantian/programs/codesnap/tests/data/fib.py:1)", "caller_lineno": 246, "ph": "X", "cat": "FEE"}, {"pid": 2783, "tid": 2783, "ts": 6556500799.241, "dur": 1.1, "name": "builtins.exec", "caller_lineno": 246, "ph": "X", "cat": "FEE"}], "displayTimeUnit": "ns", "viztracer_metadata": {"version": "0.9.5"}}
```
gaogaotiantian, line 1: I don't think you need an extra file just for the wrong extension.

tests/test_vcompressor.py
```diff
@@ -31,6 +31,62 @@ def test_basic(self):
...
+            cvf_path = os.path.join(tmpdir, "result.cvf")
+            not_exist_path = os.path.join(tmpdir, "do_not_exist.json")
+            result = self.template(
+                ["viztracer", "-o", cvf_path, "--compress", not_exist_path],
+                expected_output_file=None, success=False
+            )
+            self.assertIn("Unable to find file", result.stdout.decode("utf8"))
+
+            result = self.template(
+                ["viztracer", "-o", cvf_path, "--compress", get_json_file_path("fib")],
```
gaogaotiantian, line 45: You can use a `.py` file in `data` folder I think.

gaogaotiantian (changes requested):

tests/test_vcompressor.py
youker-Shawn: Yes, I know what you mean. That's what I wanted to write at first.
But that makes `get_json_file_path("fib.py")` look a little weird.
Maybe I could rename it to `get_tests_data_file_path`, which would be more appropriate?

PR #284, opened 2022-09-15
Added Chinese part for README

Dynesshely:


gaogaotiantian:
A Chinese README file won't help that much for understanding VizTracer. Also the way to implement it is not having two sets of docs on the same README file. Normally the way to go is to create another file and link it from the main README file.

The problem of adding this is that I will have to maintain and synchronize two sets of README files, which is not horrible but also not very beneficial.

For now, I'll turn this down. I don't think this is something I'd like to do.

Next time for this kind of change, maybe raise an issue first to discuss it before start implementing. Thanks.

Dynesshely:
I see, that's okay, next time for docs maintain, remember to @Dynesshely plz, I'd love to translate.

PR #288, opened 2022-10-01
Add instant event compresss/decompress

LorewalkerZhou:


src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -146,6 +147,57 @@ int dump_parsed_trace_events(PyObject* trace_events, FILE* fptr)
         }
     }
 
+    // Iterate through instant events
+    ppos = 0;
+    PyObject* dumps_func = NULL;
```
gaogaotiantian, line 152: Could you move the declaration on the top of the function? We should only declare variables at the top of a code block.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -146,6 +147,57 @@ int dump_parsed_trace_events(PyObject* trace_events, FILE* fptr)
...
+    PyObject* dumps_func_args = PyTuple_New(1);
+    PyObject* args_object = NULL;
+    dumps_func = PyObject_GetAttrString(json_module, "dumps");
+    if (!dumps_func) {
+        goto clean_exit;
+    }
+
+    while (PyDict_Next(instant_events, &ppos, &key, &value)) {
+        // head
+        fputc(VC_HEADER_INSTANT_EVENT, fptr);
```
gaogaotiantian, line 162: This is similar, we want to keep all the variable declarations on the top of the code block, so `fputc` should be below all the declarations.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -146,6 +147,57 @@ int dump_parsed_trace_events(PyObject* trace_events, FILE* fptr)
...
+        uint64_t tid = PyLong_AsLong(PyTuple_GetItem(key, 1));
+        const char* name = PyUnicode_AsUTF8(PyTuple_GetItem(key, 2));
+        const char* scope = PyUnicode_AsUTF8(PyTuple_GetItem(key, 3));
+        fwrite(&pid, sizeof(uint64_t), 1, fptr);
+        fwrite(&tid, sizeof(uint64_t), 1, fptr);
+        fwritestr(name, fptr);
+        fwritestr(scope, fptr);
+
+        // count
+        uint64_t ts_size = PyList_GET_SIZE(value);
```
gaogaotiantian, line 175: declaration on top also.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -146,6 +147,57 @@ int dump_parsed_trace_events(PyObject* trace_events, FILE* fptr)
...
+            // ts
+            double ts = PyFloat_AsDouble(PyList_GET_ITEM(value, 2 * idx));
+            int64_t ts64 = ts * 1000;
+            fwrite(&ts64, sizeof(int64_t), 1, fptr);
+
+            // args
+            PyTuple_SetItem(dumps_func_args, 0, PyList_GET_ITEM(value, 2 * idx + 1));
+            args_object = PyObject_CallObject(dumps_func, dumps_func_args);
+            const char* args = PyUnicode_AsUTF8(args_object);
+            fwritestr(args, fptr);
```
gaogaotiantian, line 188: I believe there's some reference bugs here. Remember that this piece of code is in a loop. For `dump_func_args`, `PyTuple_SetItem` will steal a reference that's borrowed from `value`. Because you are reusing `dump_func_args`, the previous item on the tuple will be discarded(the reference will be decreased). So we lost a reference on all the elements in `value`.

If you do want to reuse the `dump_func_args` tuple, you should probably increase the reference to all the element assigned to it. 

In a similar way, `args_object` is replaced without properly released in each loop, which leaves all the results for the function call hanging there forever.

In another matter, `args` should be defined on the top of this code block. Or you could simply skip that variable because there's no reference involved here.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -240,6 +299,39 @@ load_events_from_file(FILE* fptr)
                 }
                 Py_DECREF(name);
                 break;
+            case VC_HEADER_INSTANT_EVENT:
+                loads_func = PyObject_GetAttrString(json_module, "loads");
```
gaogaotiantian, line 303: This should not be in the loop.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -240,6 +299,39 @@ load_events_from_file(FILE* fptr)
...
+                READ_DATA(&pid, uint64_t, fptr);
+                READ_DATA(&tid, uint64_t, fptr);
+                freadstrn(buffer, STRING_BUFFER_SIZE - 1, fptr);
+                name = PyUnicode_FromString(buffer);
+                freadstrn(buffer, STRING_BUFFER_SIZE - 1, fptr);
+                scope = PyUnicode_FromString(buffer);
+                READ_DATA(&count, uint64_t, fptr);
+                for (uint64_t i = 0; i < count / 2; i++) {
+                    READ_DATA(&ts, uint64_t, fptr);
+                    freadstrn(buffer, STRING_BUFFER_SIZE - 1, fptr);
```
gaogaotiantian, line 316: There's no bound for the argument string, so we should probably use `freadstr` declared above instead of `freadstrn`.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -240,6 +299,39 @@ load_events_from_file(FILE* fptr)
...
+                    str = Py_BuildValue("s", buffer);
+                    PyTuple_SetItem(loads_func_args, 0, str);
+                    args = PyObject_CallObject(loads_func, loads_func_args);
+                    event = PyDict_New();
+                    PyDict_SetItemString(event, "pviztra    h", unicode_i);
```
gaogaotiantian, line 321: I don't understand this line..

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -240,6 +299,39 @@ load_events_from_file(FILE* fptr)
...
+                freadstrn(buffer, STRING_BUFFER_SIZE - 1, fptr);
+                name = PyUnicode_FromString(buffer);
+                freadstrn(buffer, STRING_BUFFER_SIZE - 1, fptr);
+                scope = PyUnicode_FromString(buffer);
+                READ_DATA(&count, uint64_t, fptr);
+                for (uint64_t i = 0; i < count / 2; i++) {
+                    READ_DATA(&ts, uint64_t, fptr);
+                    freadstrn(buffer, STRING_BUFFER_SIZE - 1, fptr);
+                    str = Py_BuildValue("s", buffer);
+                    PyTuple_SetItem(loads_func_args, 0, str);
```
gaogaotiantian, line 318: This is actually an interesting piece. I don't know if you considered the reference here. Each time a new `str` will be created, the reference will be stealt by `load_func_args`, and it will be discarded when the next item comes. However, in the cleanup code, you checked both `str` and `loads_func_args`, so I think there's a double release here.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -240,6 +299,39 @@ load_events_from_file(FILE* fptr)
...
+                    PyDict_SetItemString(event, "s", scope);
+                    PyDict_SetItemStringULL(event, "pid", pid);
+                    PyDict_SetItemStringULL(event, "tid", tid);
+                    PyDict_SetItemStringDouble(event, "ts", (double)ts / 1000);
+                    PyDict_SetItemString(event, "args", args);
+                    PyList_Append(trace_events, event);
+                    Py_DECREF(event);
+                    Py_DECREF(args);
+                }
+                Py_DECREF(name);
```
gaogaotiantian, line 333: You never release `scope`.

gaogaotiantian (changes requested):
I think the current code will produce some segfaults. Did you test this at all? There are multiple reference issues that I can tell.

Also, as you are developing a new feature, could you make sure at least some new tests are added to test this feature? In `test_vcompressor.py`, add some simple cases for instant event. Maybe some extra difficult ones, like if the instant argument is too long to be saved in a fixed size buffer.

LorewalkerZhou:
> I think the current code will produce some segfaults. Did you test this at all? There are multiple reference issues that I can tell.
> 
> Also, as you are developing a new feature, could you make sure at least some new tests are added to test this feature? In `test_vcompressor.py`, add some simple cases for instant event. Maybe some extra difficult ones, like if the instant argument is too long to be saved in a fixed size buffer.

Before I submit this pr, I compressed a logging_integration.json and then decompressed it. I compared several instant events in the original and decompressed files. They are consistent.

LorewalkerZhou:
I will review and modify these codes over the weekend. Thanks for the review, I'm sorry for some low-level mistakes(PyDict_SetItemString(event, "pviztra    h", unicode_i)) in my code.

Issue #285, opened 2022-09-16
Thread names with quotes and newlines cause JSON decode errors in report_builder.py:get_json


jtkouz:
Some of my threads are named with a file name and have newlines, e.g.,
`{"ph":"M","pid":23061,"tid":23060,"name":"thread_name","args":{"name":"  File "./something/otherthing.py", line 123, in exec_func_slaves
    pool = concurrency.ThreadPool(max_workers=min(num_workers, concurrency.THREAD_POOL_SIZE_MEDIUM))
_0"}}`

I worked around this by fixing up the json_str at
report_builder.py:29 for my particular data.

gaogaotiantian:
We should probably come up with a way to escape the non json compatible characters.

gaogaotiantian:
The fix is released with 0.15.5

PR #290, opened 2022-10-11
fixed escape string error and add test case

TTianshun:
this is fix for #285


gaogaotiantian:
The overall code is good. I think it might be better to have a similar name to `fprint*` (like `fprintjson`) and put `fptr` as the first argument to make it a bit more unified. Maybe something like `fprint_escape(FILE* fptr, const char* s)`.

The test is comprehensive and that's good. However, we normally try to test a feature as isolated as possible. Make the test simple and straightforward, so other people can understand the test and what it is trying to validation. Also it would be more robust, meaning it will less likely to be affected by other features.

In this case, multithreading is a good case, as thread names could potentially have any characters in it. The run function does not have to be this complicated with `sleep`, as we are only interested in the thread names. Maybe having 4 threads is not necessary either.

Magic comment is a cool and more fragile feature. I think we could simply do

```python
tracer = get_tracer()
tracer.log_instant()
```

to avoid noises from potential magic comment issues. It's also more friendly to people who are not familiar with the magic comments.

Also, we probably do not need a separate file just for this test. I'd normally put this kind of test into `test_regression.py`, where all the regression tests are kept. You can find another file that you think suitable to put it in, but a new file is a bit too much.

You don't need to check `stdout` as the execution would fail if the fix was not in. `expected_output_file` is probably default to `result.json`? It's not a big deal to explicitly put it there (some of my tests do), but it's also okay to omit that.

gaogaotiantian:
So, a very important step to write regression test is to - make sure it FAILS without your changes. Normally we write a regression test first, make sure it fails, then fix the bug and see if the tests passes.

If you have done that then you probably will find that your test runs fine without your changes. The strings you are testing are `\\\"`(as an example), which is `'\'`and `'"'`. So this works fine with the current code as it would be a valid json.

Instead, we should test against `\"`, which would be a single `'"'` without the fix and break the code.

I checked the code and as the `exit_rountine` runs in `finalizers`, the exceptions are caught. It's okay to use stdout to determine the result of the test.

TTianshun:
I test it on the `master` branch, and it failed. 
image
I write the srting like `\\\"` because there's string escape while writing the test code into cmdline_test.py file. And in cmdline_test.py, it would be `\"`.
Here's cmdline_test.py that the code generated:
```
import threading
from viztracer import get_tracer, VizCounter, VizObject


def fib(n):
    if n < 2:
        return 1
    return fib(n - 1) + fib(n - 2)


class MyThread(threading.Thread):
    def run(self):
        fib(7)


tracer = get_tracer()

# test object event name escape with and without args
obj = VizObject(tracer, "test \\ \" \b \f \n \r \t")
obj.test = "test \\ \" \b \f \n \r \t"

# test counter event name escape with and without args
counter = VizCounter(tracer, "test \\ \" \b \f \n \r \t")
counter.test = 10

# test instant event name escape with and without args
tracer.log_instant("test \\ \" \b \f \n \r \t")
tracer.log_instant("test \\ \" \b \f \n \r \t", "test \\ \" \b \f \n \r \t")

# test thread name escape
test_thread = MyThread(name = "test \\ \" \b \f \n \r \t")
test_thread.start()
test_thread.join()
```

gaogaotiantian:
Ah, you are right. I directly copied the code into a new py file and tested. Yes it should be double escaped as it's in a big string.

PR #305, opened 2022-12-17
Add compressing mechamism for all other non-frequent events

TTianshun:
This is compressor for instant events.

I've learned about compress recently. Actually, we did not apply any algorithm on the compressor. What we are doing is to rearrange to reduce the redundancy of the data. And it's helpful for compressing.
I did some simple tests, here's the result:
I tested `different_sorts.py` with `file_info=False` and used different algorithms to compress `result.json` and `result.cvf`, here's the results:
```
zlib compress result.json, origin size is 3563190, compressed size is 127722
zlib compress result.cvf, origin size is 448242, compressed size is 97918
lzma compress result.json, origin size is 3563190, compressed size is 83900
lzma compress result.cvf, origin size is 448242, compressed size is 69148
```


gaogaotiantian:
Transforming the format of the data is a kind of compressing. Compressing by definition is just making it smaller. It's just there are generic ways to compress an arbitrary binary file. It would be interesting to check if a generic compressing algorithm like `zlib` or `lzma` can furthur compress the `cvf` file when the original file is **large** and filled with almost all **fee**s, as that's the most common case in reality.

So to decide whether we should apply an extra layer of compression(or maybe our current compressing algorithm is so bad that it's just not worth it), we should try to create a large file(>100M?) with maybe just `fib()` and see the outcomes of different methods. It's possible that the generic compressing algorithm actually performs much better than I expected.

src/viztracer/modules/vcompressor/vc_dump.c
```diff
@@ -161,14 +162,169 @@ int dump_parsed_trace_events(PyObject* trace_events, FILE* fptr)
...
+    }
+clean_exit:
+    if (PyErr_Occurred()) {
+        return 1;
+    }
+
+    return 0;
+}
+
+int write_instant_args(PyObject* instant_args, FILE* fptr) {
```
gaogaotiantian, line 190: The major way we "compress" data in `vcompressor` is to reduce the redundancy, especially when multiple data entries have the same info except for ts. However, for instant events, it seems like we can't do that easily because each instant event could potentially have completely different data. In that case, the method to rearrange data like this would only save us the redundant fields like `ph` or `cat`, not the real data.

So, to make this simpler, considering the manual rearrangement of data was not that helpful, maybe just dump the whole thing as binary? For each instant data with the same pid&tid&name&scope, dump the dict you kept as json(plus zlib?), the code should be simpler(no loops in `write_instant_args` function anymore) and the compression rate should be better(with zlib of course).

The code to do it is basically in `dump_file_info`. Most of the code in that function is to make `zlib` + `json` work. I'm thinking maybe we could separate the `zlib` + `json` part out as an utility function - which could be called from both dumping file info and dumping instant events.

The good thing about this is that we can keep one less compressing method - the current way you compress instant events is unique and probably won't be applicable to other cases(or for the cases it's applicable, they could/should be replaced with `json` + `zlib`).

gaogaotiantian (changes requested):

TTianshun:
> Transforming the format of the data is a kind of compressing. Compressing by definition is just making it smaller. It's just there are generic ways to compress an arbitrary binary file. It would be interesting to check if a generic compressing algorithm like `zlib` or `lzma` can furthur compress the `cvf` file when the original file is **large** and filled with almost all **fee**s, as that's the most common case in reality.
> 
> So to decide whether we should apply an extra layer of compression(or maybe our current compressing algorithm is so bad that it's just not worth it), we should try to create a large file(>100M?) with maybe just `fib()` and see the outcomes of different methods. It's possible that the generic compressing algorithm actually performs much better than I expected.

Sorry for the late response because I had COVID-19 days ago. 

I think maybe we need more tests on different kinds of json file. `Zlib` compression may be effective for `fib()` because  it has a lot of repetitive args. But for those programs which are not as organized as `fib()`. the compression may be not so effective. I'll test it later.

As for "separate the `zlib` + `json` part out as an utility function", maybe we can directly put `file_info` and `instant event` in `cvf` file and compress the whole `cvf` file. Of course a further test needed to be done.

gaogaotiantian:
> 
> Sorry for the late response because I had COVID-19 days ago.
> 
> I think maybe we need more tests on different kinds of json file. `Zlib` compression may be effective for `fib()` because it has a lot of repetitive args. But for those programs which are not as organized as `fib()`. the compression may be not so effective. I'll test it later.
> 
> As for "separate the `zlib` + `json` part out as an utility function", maybe we can directly put `file_info` and `instant event` in `cvf` file and compress the whole `cvf` file. Of course a further test needed to be done.

Hope you are feeling better now. Actually, `fib` would be the one that the current compressing algorithm performs well at. The current way basically elimiates the extra information on the same function by only recording the timestamps. However, it would be nice to have a more real-life benchmark to test it out.

If eventually we decided to just compress the whole cvf file(or rather, compress the file to be the final cvf file), then sure we can skip the zlib part. However, we still need some mechanism to label the "json" part inside a binary file.

TTianshun:
Here's my test, one is fib, and another is a python game using pygame. As I know, many games was developed using python.
| Function    | Compress Method | Uncompressed Size | Compressed Size | Compress Time | Compress Rate |
| ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |
| fib  | zlib | 160314109 | 4329927 | 0.8s | 2.7% |
| fib  | lzma | 160314109 | 3392560 | 7.1s | 2.1% |
| fib  | vcompressor | 160314109 | 16000187 |0.8s | 10.0% |
| fib  | vcompressor + zlib | 160314109 | 3808128 | 1.2s | 2.4% |
| fib  | vcompressor + lzma | 160314109 | 2612632 | 8.9s | 1.6% |
| game | zlib | 122974645 | 6410428 | 0.9s | 5.2% |
| game | lzma | 122974645 | 4335556 | 9.9s | 3.5% |
| game | vcompressor | 122974645 | 13771658 | 0.8s | 11.2% |
| game | vcompressor + zlib | 122974645 | 3756816 | 1.1s | 3.1% |
| game | vcompressor + lzma | 122974645 | 2580504 | 7.7s | 2.1% |

As we can see, compress the whole `cvf` file are better in compress rate. The traditional compress functions perform better on well ordered file such as `fib`. In `fib` situation, compressing the whole `cvf` seemed to make less optimization. But in complex situation, the optimization seems to be obvious.

TTianshun:
the data only contains `FEE` in `TraceEvents`, and `file_info` is also not contained

gaogaotiantian:
Interesting. So for `fib`, vcompressor is about 4x in size and for game it's about 2x. I think now is the time to slow down the other features and see if we could do better than that. If vcompressor can only do 2x-4x compared to zlib, then we should simply use zlib directly.

The first idea comes to mind is to compress timestamp better. We are using an 8 byte integer for a single timestamp, which takes most of the disk spaces for a trace report. We should try to do a "diff-based" solution. Where the "next" timestamp only records the diff from the previous one. We also need to employ variant integer encoding - to use different size of bytes to record integers.

There are a couple of ways to implement it:
1. Use the most significant bit to indicate whether there are more bytes to follow. The bit wasted for non-data will be 1/8, but for any diff that's less than or equal to 7 bit, it will only require 1 byte.
2. Use the first 3 bits to indicate how many bytes will we use for the timestamp. The bit wasted is fixed, but we can only deal with 5-bit difference with one byte.

One thing to notice is how we handle our timestamp - the minimum resolution is about 20ns and we could probably do something on that end as well.

TTianshun:
> Interesting. So for `fib`, vcompressor is about 4x in size and for game it's about 2x. I think now is the time to slow down the other features and see if we could do better than that. If vcompressor can only do 2x-4x compared to zlib, then we should simply use zlib directly.
> 
> The first idea comes to mind is to compress timestamp better. We are using an 8 byte integer for a single timestamp, which takes most of the disk spaces for a trace report. We should try to do a "diff-based" solution. Where the "next" timestamp only records the diff from the previous one. We also need to employ variant integer encoding - to use different size of bytes to record integers.
> 
> There are a couple of ways to implement it:
> 
> 1. Use the most significant bit to indicate whether there are more bytes to follow. The bit wasted for non-data will be 1/8, but for any diff that's less than or equal to 7 bit, it will only require 1 byte.
> 2. Use the first 3 bits to indicate how many bytes will we use for the timestamp. The bit wasted is fixed, but we can only deal with 5-bit difference with one byte.
> 
> One thing to notice is how we handle our timestamp - the minimum resolution is about 20ns and we could probably do something on that end as well.

I read the code about how to get timestamp. The minium resolution is about 20ns, maybe we can directly divide timestamp by 20ns. In my mind, the decompressed timestamp maybe different but the result of vizviewer shoule be the same with original file.

gaogaotiantian:
@TTianshun Let's do the timestamp change first and circle back to this PR.

TTianshun:
> @TTianshun Let's do the timestamp change first and circle back to this PR.

OK, I'll finish that change first.

TTianshun:
It occurred to me that we can put all the args in a list then use json dumps and compress. It may be better in compression rate.

gaogaotiantian:
I took a look at the code. Abstracting the common code to "dump a json(dict)" is definitely helpful. However, we need to have an engineering mind here - that instant event will never be the majority of the trace file, its compression rate does not matter. The only thing we should do about instant events, is to make it simple and correct, until the day comes (if it ever does) when this is an important feature and more people are heavily relying on it.

So, here's what I propose - just put all the instant events in a list, and json+zlib it. Don't even bother to categorize them based on `(pid, tid, name, scope)`. Most of the duplicated string piece in json should be taken care of by `zlib` anyway. This would make our code much more maintainable and it won't hurt in actual cases.

Of course, if you are interested, you can compare the compression ratio, I doubt it would make a big difference.

TTianshun:
> I took a look at the code. Abstracting the common code to "dump a json(dict)" is definitely helpful. However, we need to have an engineering mind here - that instant event will never be the majority of the trace file, its compression rate does not matter. The only thing we should do about instant events, is to make it simple and correct, until the day comes (if it ever does) when this is an important feature and more people are heavily relying on it.
> 
> So, here's what I propose - just put all the instant events in a list, and json+zlib it. Don't even bother to categorize them based on `(pid, tid, name, scope)`. Most of the duplicated string piece in json should be taken care of by `zlib` anyway. This would make our code much more maintainable and it won't hurt in actual cases.
> 
> Of course, if you are interested, you can compare the compression ratio, I doubt it would make a big difference.

  I agree with that. Besides instant events, there's other events such as VizObject and maybe custom events. If we compress them seperatly, the code will be much more complex. So I directly dumps and compress the rest events.

gaogaotiantian:
Compressing all the "other" events is a great idea. They will be only a very small part of the trace and it should not take us too much time to take care of them. The code looks good to me. Just a couple of thoughts:

1. Maybe update `README.md` file for our compressing strategy?
2. Could you write some basic tests for the "other" events that you found? That could give us some basic coverage.

TTianshun:
It seemes that in some case, the parsed timestamp can have about 10 ns difference due to some operation on double type. But it should be ok I guess. For now I haven't seen a very big deviation.

gaogaotiantian:
> It seemes that in some case, the parsed timestamp can have about 10 ns difference due to some operation on double type. But it should be ok I guess. For now I haven't seen a very big deviation.

We don't want an occasionally failing test. If we believe this deviation is acceptable and unavoidable, we should relax the restrictions - make it 0.11 ns for example.

gaogaotiantian:
Seems like there's another 0.10 left? The new test you created. Should that be modified as well?

TTianshun:
I think this deviation comes from two aspects. 
The first is about this https://github.com/gaogaotiantian/viztracer/pull/289#issuecomment-1270563430 we have met before. In `FEE` the timestamp will be multipled by 100 when compressing and be divided by 100 when decompressing. So the max deviation for this situation is 0.01.
The second is that viztracer will keep three decimal places in `result.json`, if a timestamp is 1.009, then multiple it by 100 and convert it to `int` will cause deviation of 0.009.
So I think maybe 0.02 is reasonable and won't cause failed test.
For `FEE` tests, I will change it to 0.02. For the tests of other events, 0.01 is enough. But i think we can change it to 0.02 to make the tests consistent.

gaogaotiantian:
Hmm, 0.02 is a dangerous line because that's our artificial increment for minimum timer resolution. If the deviation is greater than 0.02 then it means the order of events could be potentially swapped. Let's do `0.011` for all the test and see if it would fail frequently in nightly validation. If so, we need to do something to address the issue.

TTianshun:
> Hmm, 0.02 is a dangerous line because that's our artificial increment for minimum timer resolution. If the deviation is greater than 0.02 then it means the order of events could be potentially swapped. Let's do `0.011` for all the test and see if it would fail frequently in nightly validation. If so, we need to do something to address the issue.

Fixed it!

gaogaotiantian:
Congrats! This is a large PR and after this, I think the trace-log-compressor is ready to go into the master branch.

TTianshun:
> Congrats! This is a large PR and after this, I think the trace-log-compressor is ready to go into the master branch.

I think maybe we need another pr for args in FEE, and it will be ready to go.

Issue #318, opened 2023-03-23
Test failure - test_cellmagic (tests.test_jupyter.TestJupyter)


gaogaotiantian:
This probably because a library is updated. `ipywidgets` is a potential candidate.

@TTianshun are you interested in this bug? I can work on it when I have some time. It just seems like a good opportunity to get in touch with some first-hand bugs in viztracer.

TTianshun:
OK, I'll try to fix it!

PR #319, opened 2023-03-25
Fix 318

TTianshun:
There's several things about this issue #318 
1. The test fail: I think the test fail is because some bugs of ipywidgets, https://github.com/jupyter-widgets/ipywidgets/issues/3739 and https://github.com/jupyter-widgets/pythreejs/issues/399 are related bugs. So to fix the test fail, for now we can use `ipywidgets==8.0.4`
2.  The skipped test: I found that this fail was only occurred in python 3.7 because of IPython. IPython 8+ supports Python 3.8 and above, and Python 3.7 was supported with 7.x. for Python 3.8 and above, `from IPython.testing.globalipapp import start_ipython` will import `pytest` as dependency, which is not installed and triggered `ImportError`. One solution is to add `pytest` to `requirements-dev.txt`. Another solution is to use this samplecode to test:
  ```python
  from traitlets.config.loader import Config
  import tempfile
  from pathlib import Path
  
  def default_config():
      config = Config()
      config.TerminalInteractiveShell.colors = 'NoColor'
      config.TerminalTerminalInteractiveShell.term_title = False,
      config.TerminalInteractiveShell.autocall = 0
      f = tempfile.NamedTemporaryFile(suffix=u'test_hist.sqlite', delete=False)
      config.HistoryManager.hist_file = Path(f.name)
      f.close()
      config.HistoryManager.db_cache_size = 10000
      return config
  
  def get_ipython():
      from IPython.terminal.interactiveshell import TerminalInteractiveShell
      if TerminalInteractiveShell._instance:
          return TerminalInteractiveShell.instance()
      config = default_config()
      config.TerminalInteractiveShell.simple_prompt = True
      shell = TerminalInteractiveShell.instance(config=config)
      return shell
  
  ip = get_ipython()
  ip.magic("load_ext viztracer")
  ip.run_cell_magic(magic_name="viztracer", line="", cell="print(1)")
  ```
3. I added `ipykernel` because of this issue https://github.com/jupyter-widgets/ipywidgets/issues/3731. For now maybe we don't need this. But if we need to use a newer version of ipywidgets, this is necessary.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,6 +9,6 @@ mypy
 coverage
 
 # 3rd party packages for test
-ipywidgets
+ipywidgets==8.0.4
 loky>=3.0.0
 jaxlib; sys_platform != 'win32' and python_version < '3.11'
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,6 +9,8 @@ mypy
 coverage
 
 # 3rd party packages for test
+ipykernel
+pytest
 ipywidgets==8.0.4
 loky>=3.0.0
 jaxlib; sys_platform != 'win32' and python_version < '3.11'


gaogaotiantian:
I'm okay with the temporary `ipywidgets` version freeze, preferably with a comment in `requirements-dev.txt` file so that we can remove it later when the bug is fixed.

`ipykernel` should not be included now, we should include that when the bug is fixed. We can leave a TODO in the issue so that we won't forget.

The part I don't understand is why we need `pytest`. First of all, which is the skipped test you mentioned? Why didn't I have this issue before this bug was introduced? Are you saying somehow the library requires `pytest` to run? I think that's a problem. Can you be more specific about the `pytest` thing?

TTianshun:
the skipped test is `tests.test_jupyter.TestJupyter`, for example:

TTianshun:
If i directly run the sample code:
```
from IPython.testing.globalipapp import start_ipython
ip = start_ipython()
ip.magic("load_ext viztracer")
ip.run_cell_magic(magic_name="viztracer", line="", cell="print(1)")
```
the output is:
```
Traceback (most recent call last):
  File "/home/pts/viztracer_fix318/viztracer/_test_jupyter.py", line 32, in <module>
    from IPython.testing.globalipapp import start_ipython
  File "/home/pts/viztracer_fix318/viztracer/venv-3.11/lib/python3.11/site-packages/IPython/testing/globalipapp.py", line 18, in <module>
    from . import tools
  File "/home/pts/viztracer_fix318/viztracer/venv-3.11/lib/python3.11/site-packages/IPython/testing/tools.py", line 30, in <module>
    from . import decorators as dec
  File "/home/pts/viztracer_fix318/viztracer/venv-3.11/lib/python3.11/site-packages/IPython/testing/decorators.py", line 142, in <module>
    skip_win32 = skipif(sys.platform == 'win32',
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pts/viztracer_fix318/viztracer/venv-3.11/lib/python3.11/site-packages/IPython/testing/decorators.py", line 88, in skipif
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

gaogaotiantian:
Okay I understood now. So `IPython.testing.globalipapp` is not an official documented feature. Then let's do a more stable way to get the interactive shell to test. Could you take a look at the current code you provided for the shell and see if it could be trimmed down a bit? For example, is the history necessary? Can we minimize it for our testing purpose(mainly to avoid fixing problems in the future)?

TTianshun:
I made some changes according to this:https://github.com/ipython/ipython/issues/12334, and the code works.

requirements-dev.txt
```diff
@@ -9,6 +9,6 @@ mypy
 coverage
 
 # 3rd party packages for test
-ipywidgets
+ipywidgets==8.0.4
```
gaogaotiantian, line 12: Can we add a comment here about this is temporary and should be unfreezed as long as ipywidgets is fixed?

tests/test_jupyter.py
```diff
@@ -10,14 +10,13 @@ class TestJupyter(BaseTmpl):
     def setUp(self):
         super().setUp()
         try:
-            from IPython.testing.globalipapp import start_ipython
-            self.start_ipython = start_ipython
+            from IPython.terminal.interactiveshell import TerminalInteractiveShell
+            self.ip = TerminalInteractiveShell.instance()
```
gaogaotiantian, line 14: If we have more tests in the future, are we supposed to use this specific instance repeatedly, or we should create a new one? If we are supposed to create a new one, then maybe do not store the instance in the test class, store the create method?

gaogaotiantian (changes requested):

PR #326, opened 2023-04-07
Compress fee args

TTianshun:
Add compressor for fee args, which will happen when enable `log_func_args` and `log_func_retval`

### DIFF
--- a/src/viztracer/modules/vcompressor/README.md
+++ b/src/viztracer/modules/vcompressor/README.md
@@ -40,7 +40,7 @@ header(header) - pid(pid) - tid(tid) - name(str)
 
 ### FEE
 
-header(header) - pid(pid) - tid(tid) - name(str) - count(uint64) - [start(ts) - dur(ts)]*
+header(header) - pid(pid) - tid(tid) - name(str) - count(uint64) - args_flag - args offset(if there's args) - [start(ts) - dur(ts)]* - args list (if there's args)
 
 ### File info
 
--- a/src/viztracer/modules/vcompressor/README.md
+++ b/src/viztracer/modules/vcompressor/README.md
@@ -40,7 +40,7 @@ header(header) - pid(pid) - tid(tid) - name(str)
 
 ### FEE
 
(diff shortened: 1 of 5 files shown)


gaogaotiantian:
If we put the argument exist flag after the timestamps, we can get rid of the offset right? And the logic would be simpler.

TTianshun:
> If we put the argument exist flag after the timestamps, we can get rid of the offset right? And the logic would be simpler.

I write like this because I want to iterate the list only once while reading and writing. if we put the argument exist flag after the timestamp. We have to iterate the fee events again. considering fee is the major of trace file, the time cost might be more.

gaogaotiantian:
There's a more critical issue - you assumed that all the FEEs of the same function will either have `args` or not. That's not always the case. It's possible that some of the FEEs have `args` and the others don't. We need to take care of that.

TTianshun:
> There's a more critical issue - you assumed that all the FEEs of the same function will either have `args` or not. That's not always the case. It's possible that some of the FEEs have `args` and the others don't. We need to take care of that.

That should be considered in `parse_trace_events` function. For the same function events with or without args, they are seperately stored and wrote. The test case should also include it.

gaogaotiantian:
> > There's a more critical issue - you assumed that all the FEEs of the same function will either have `args` or not. That's not always the case. It's possible that some of the FEEs have `args` and the others don't. We need to take care of that.
> 
> That should be considered in `parse_trace_events` function. For the same function events with or without args, they are seperately stored and wrote. The test case should also include it.

Ah, right, you kept them into different keys.

gaogaotiantian:
Then let's get rid of the flag piece. Only have the offset - if the offset is `0` then it means no args. It's impossible for the offset to be `0` as long as there are entries. The logic would still be simpler.

For writing it's actually fine - you are assembling a list then write to the file anyway. For reading yeah it's probably easier to have both the timestamp and arg list ready to generate events.

TTianshun:
Could you review these changes? @gaogaotiantian

PR #328, opened 2023-04-10
Better popen patching

mwerschy:
`viz_args` needs to be copied before modifying it to avoid accumulating args across invocations. Can also handle combined options e.g. `-Es` now.

### DIFF
--- a/src/viztracer/patch.py
+++ b/src/viztracer/patch.py
@@ -3,6 +3,7 @@
 
 
 import functools
+import itertools
 from multiprocessing import Process
 import os
 import re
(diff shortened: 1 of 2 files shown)


gaogaotiantian:
That's good improvements. Actually not being able to handle combined options and having duplicate args are kind of bugs. Are you willing to write some regression tests on this? If not, I can work on it later. Could you add some comment at copying `viz_args` mentioning the reason of it? Thanks.

gaogaotiantian:
Yes, removing those options is required - but that's not a super common bug probably because people are not using combined options that much(at least with viztracer). So this is a bug - which we should prevent from happening again in the future with regression tests :)

gaogaotiantian:
Passing them through won't work because you are not passing them to Python. After you patch the subprocess, you'll pass them to `viztracer`, which is not able to recognize those options. How did you pass them through? The current PR still throw out all the options(it does a better job, throwing out combined options).

gaogaotiantian:
Actually, if we do `python -m viztracer`, we can probably pass through those options fine.

gaogaotiantian:
I just realized that it's possible that the command itself has command line arguments like `python myscript.py -I` and the current methods will strip the arguments for the script. We can pass through the options for python but we need to tread carefully. If you still want to work on this, it's probably be more complicated (with regression tests). Otherwise you can monkey patch the code you are using now and I'll try fo fix this later.

gaogaotiantian:
Well the tests all passed and your new code got a 100% coverage so that's very good. The actual case viztracer is dealing with will not be so complicated, but the code quality is good and the test is comprehensive so I'll trust that. Bottom line is - what's working before is still working now, that's very important. And I do believe you solved a LOT of corner cases so thanks for that! As long as the lint passed I'll accept the PR.

gaogaotiantian:
Thank you for the contribution!

PR #331, opened 2023-04-16
Trace log compressor docs

TTianshun:
I've tried to write the basic usage docs for `VCompressor`. My English is poor. If the docs are not good,  you can close this pr.

### DIFF
--- a/docs/source/basic_usage.rst
+++ b/docs/source/basic_usage.rst
@@ -258,16 +258,16 @@ For detailed commands, please refer to :doc:`virtual_debug`
 Compress Your Report
 --------------------
 
-VizTracer supports compressing your json report. The general compress ratio is about 1% - 3% for a large report.
+VizTracer supports compressing your json report. The general compression ratio is about 50:1 to 100:1 for a large report.
 
 You can compress your report with ``--compress``.
 
 .. code-block:: 
 
-    viztracer -o result.cvf --compress result.json
+    viztracer --compress result.json -o result.cvf 
 
 You can also decompress your report with ``--decompress``
 
 .. code-block:: 
 
-    viztracer -o result.json --decompress result.cvf
+    viztracer --decompress result.cvf -o result.json 
--- a/docs/source/basic_usage.rst
+++ b/docs/source/basic_usage.rst
@@ -258,7 +258,7 @@ For detailed commands, please refer to :doc:`virtual_debug`
 Compress Your Report
 --------------------
 
-VizTracer supports compressing your json report. The compress rate is better than commonly used compression methods like ``zlib`` and ``lzma``.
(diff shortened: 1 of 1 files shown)


docs/source/basic_usage.rst
```diff
@@ -254,3 +254,20 @@ interact with it. Even better, you can **go back in time** because you know what
     vdb <your_json_report>
 
 For detailed commands, please refer to :doc:`virtual_debug`
+
+Compress Your Report
+--------------------
+
+VizTracer supports compressing your json report. The compress rate is better than commonly used compression methods like ``zlib`` and ``lzma``.
```
gaogaotiantian, line 261: The documentation does not need to say it's better than the common methods. However, we can give a general compress rate here.

gaogaotiantian (changes requested):
Just one comment and it should be fine.

docs/source/basic_usage.rst
TTianshun: updated here!

docs/source/basic_usage.rst
```diff
@@ -254,3 +254,20 @@ interact with it. Even better, you can **go back in time** because you know what
     vdb <your_json_report>
 
 For detailed commands, please refer to :doc:`virtual_debug`
+
+Compress Your Report
+--------------------
+
+VizTracer supports compressing your json report. The general compress ratio is about 1% - 3% for a large report.
```
gaogaotiantian, line 261: I think the compression ratio is defined as `uncompressed size / compressed size` [wiki](https://en.wikipedia.org/wiki/Data_compression_ratio). Here we could say 50:1 to 100:1 maybe.

docs/source/basic_usage.rst
```diff
@@ -254,3 +254,20 @@ interact with it. Even better, you can **go back in time** because you know what
...
+
+You can compress your report with ``--compress``.
+
+.. code-block:: 
+
+    viztracer -o result.cvf --compress result.json
```
gaogaotiantian, line 267: I think maybe ``viztracer --compress result.json -o result.cvf`` makes more sense to the users. Compress then set the output file, compared to set the output file then compress.

docs/source/basic_usage.rst
```diff
@@ -254,3 +254,20 @@ interact with it. Even better, you can **go back in time** because you know what
...
+
+You can also decompress your report with ``--decompress``
+
+.. code-block:: 
+
+    viztracer -o result.json --decompress result.cvf
```
gaogaotiantian, line 273: Same thing here for the order.

gaogaotiantian (changes requested):
Compression ratio and one tiny order thing I noticed

TTianshun:
> Compression ratio and one tiny order thing I noticed

Those advices sound reasonable. I've changed that.

PR #348, opened 2023-06-15
fix pid_suffix and fix test case

### LINKED ISSUE #346: Please verify that the --pid_suffix option works as intended
### Discussed in https://github.com/gaogaotiantian/viztracer/discussions/345

<div type='discussions-op-text'>

<sup>Originally posted by **kllp** June 13, 2023</sup>
Recent versions of viztracer don't seem to append the PID to the filename. Instead it creates `result.json`.

`viztracer --pid_suffix single_process.py` appends the PID in version 0.12.3 but not in 0.15.6

https://viztracer.readthedocs.io/en/latest/concurrency.html#combine-reports</div>

TTianshun:
Resolve #346

### DIFF
--- a/src/viztracer/main.py
+++ b/src/viztracer/main.py
@@ -580,7 +580,6 @@ class VizUI:
             ofile = prefix_pid + suffix
         else:
             ofile = self.ofile
-        print("ofile", ofile)
 
         self.wait_children_finish()
         builder = ReportBuilder(
--- a/src/viztracer/main.py
+++ b/src/viztracer/main.py
@@ -213,7 +213,7 @@ class VizUI:
         if options.output_file:
             self.ofile = options.output_file
         elif options.pid_suffix:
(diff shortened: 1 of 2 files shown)


gaogaotiantian:
Hmm, this is still not the expected behavior. If the user specifies ``-o output.json --pid_suffix``, the output file should be ``output_<pid>.json``. Also there could be potentially other formats in the future(html is kind of deprecated now). I think we should deal with it in `VizUI.save`.

TTianshun:
> Hmm, this is still not the expected behavior. If the user specifies `-o output.json --pid_suffix`, the output file should be `output_<pid>.json`. Also there could be potentially other formats in the future(html is kind of deprecated now). I think we should deal with it in `VizUI.save`.

Fixed it. Actually I'm not sure if a user specific `-o` and `--pid_suffix` in a commandline, what kind of result he expects.

PR #356, opened 2023-08-01
Small change：add a variable to  function "_generate_test_data" to make it easier to read

monkeywq:
Hello，I review  the function _generate_test_data in  test_vcompressor.py. I think it is better to add a variable to replace extern function get_tests_data_file_path. After that the code become easier to read and reduce the number of function calls.
its  my first time to create pull request.Please tell me how to correct it if  I make many mistakes，Thanks！


gaogaotiantian:
Your PR contains 32 commits to 80+ files. Also `trace_log_compressor` has been merged already so it should not be the target.

PR #360, opened 2023-08-04
Updates for @log_sparse with context manager and parameters for %%viztracer

### LINKED ISSUE #359: `log_sparse=True` doesn't work with context manager
Script `sparse.py`:
```
import time
from viztracer import log_sparse, VizTracer

def f():
    time.sleep(0.1)

@log_sparse
def g():
    time.sleep(0.1)

with VizTracer(output_file="result_inner.json", log_sparse=True) as tracer:
    f(); g()
```
Output of `python sparse.py; vizviewer result_inner.json`:
image

Script `sparse_2.py`:
```
import time
from viztracer import log_sparse, VizTracer

def f():
    time.sleep(0.1)

@log_sparse
def g():
    time.sleep(0.1)

f(); g()
```
Output of `viztracer --log_sparse sparse_2.py; vizviewer result.json`:
image

Versions (`python -V; pip freeze`):
```
Python 3.11.2
objprint==0.2.2
viztracer==0.15.6
```

AlexeyKozhevin:
- Closes #359 
- Also allows to define functions decorated by `@log_sparse` outside of context manager
- Add parameters for `%%viztracer`
- Extend tests
- Add `shield_ignore` method to reset counter of the ignored stack calls


src/viztracer/cellmagic.py
```diff
@@ -32,6 +36,40 @@ def view():  # pragma: no cover
 
             display(button)
 
+        def make_kwargs(self, line, tracer_signature, viewer_signature):
```
gaogaotiantian, line 39: You basically wrote an argument parser here, which would be difficult to maintain and requires a lot of tests (viztracer keeps a 100% coverage). I don't think this should be the way to go, especially considering there could be conflict arguments (for example, eliminating `--once` is a very bad idea for the viewer).

I don't want the magic cells to have full flexibility as the CLI, that could be unnecessarily difficult to maintain. However, having some options like ``--port`` might be beneficial. Still, I prefer a white list way instead of this auto-importing kind of way (this implementation will introduce a silent dependency that VizTracer and ServerThead can't share args with the same name, which is not ideal).

To be honest, cell magic is not the top-priority for viztracer - it would be nice if it works, but it's totally fine that it won't work under some circumstances. To leverage its full capability, the users can always use the inline `VizTracer` context and open the report with code.

So, I think I'd prefer this piece to be removed, and if you want some useful argument for the cell magic, use something like `IPython.core.magic_arguments`, which we do not need to test for its validity.

src/viztracer/decorator.py
```diff
@@ -97,7 +106,16 @@ def wrapper(*args, **kwargs) -> Any:
...
                 "cat": "FEE",
             }
-            tracer.add_raw(raw_data)
+            local_tracer.add_raw(raw_data)
             return ret
 
         return wrapper
+
+
+def _log_sparse_wrapper_with_args(stack_depth: int = 0, dynamic_tracer_check: bool = False) -> Callable:
```
gaogaotiantian, line 115: This function is only used once and the content is a single line. I don't see a very strong reason to abstract it out. I think we can just return the `partial` when the `func` is `None` in `log_sparse`.

gaogaotiantian (changes requested):
Overall the docs and the code is good. I have left some comments for the part that I personally have some concerns. Also as I pursue 100% coverage for viztracer, there may or may not be some extra tests requirement - let's see the coverage result.

src/viztracer/decorator.py
```diff
@@ -57,47 +57,83 @@ def wrapper(*args, **kwargs) -> Any:
...
-            return lambda f: f
-        return func
+def _log_sparse_wrapper(func: Optional[Callable], stack_depth: int = 0,
+                        dynamic_tracer_check: bool = False) -> Callable:
+    if not dynamic_tracer_check:
+        tracer = get_tracer()
+        if tracer is None or not tracer.log_sparse:
+            return func
+
+    if stack_depth > 0:
```
gaogaotiantian, line 67: Wow this part is getting out of hand fast :)

So here's my suggestion:
1. We used to have two wrappers depending on whether `func` is `None - that's because the layers are different. Now that we are using `partial`, we can use just one `wrapper` and check `stack_depth` when we need to log the trace. Yes it would introduce an extra run-time if statement, but the overhead is already large in this case. We could reduce plenty of redundant code.
2. The if logic below is getting crazy - even I would have a very difficult time following it. We should try to sort out the logic and simplify that. When we are on the if statement, we basically only have two options:
    * Execute the function normally but make sure VizTracer does not ignore it
    * Do a manual sparse log, but two methods based on `stack_depth`
3. We should only do the latter when `log_sparse==True` and `tracer.enable==False`, for all other cases, we should do the former.
4. It's okay to have a private method of tracer to `setignorestackcounter`, but can we do a public method of VizTracer like `shield_ignore()` which is a context so we can do
    ```python
    with tracer.shield_ignore():
        func()
    ```
    This requires some count manipulation for the context method (because the counter would be changing going inside and outside of the method), but it would be a one time thing and we can use that later.
5. `shield_ignore()` should be a no-op when the tracer is disabled (stopped) or when the tracer is not in ignore state (I think it should just work if the logic is implemented correctly), so we do not even need to have a separate case for "normal execution" (your current `else` case).

In this way, I believe we can greatly reduce the complexity (both length and logic) of the decorator and make this much easier to read and understand.

gaogaotiantian (changes requested):
The current logic has became a bit hard to follow. I made a structural suggestion for the current logic, feel free to discuss it if you disagree.

gaogaotiantian:
Hi @AlexeyKozhevin , are you still interested in working on this PR? I can take over if you think this is too much work. I think the PR still has its value and I'd like to merge it once it's polished.

AlexeyKozhevin:
Hi, unfortunately, I haven't had enough time lately, but I expect to get back to the task in the next two weeks. If anything changes, I will post. Thank you very much for your interest in PR

gaogaotiantian:
> Hi, unfortunately, I haven't had enough time lately, but I expect to get back to the task in the next two weeks. If anything changes, I will post. Thank you very much for your interest in PR

Sounds good! Thanks for your time. Python 3.12.0 is coming soon so I plan to update viztracer to 0.16.0 (there are a lot of payloads for the version now) and support 3.12.

gaogaotiantian (changes requested):
Thanks for working on this.

Seems like you forgot to add the newly added file? I'm not sure if we want a new file and a new class. Is it possible to implement the shield logic in the function directly with `contextmanager`? I would expect the logic to be rather straightforward.

Also, you made it possible to pass arguments to `%%viztracer`, which is great. The user does not know that unfortunately. Could you add something in the docs, after the jupyter part, just mention that some of the command line arguments are supported, and list the supported arguments? Thanks!

Other than that, I think the current code looks good.

gaogaotiantian:
> Sorry, I really forgot one file. Thank you for comments, I've updated `basic_usage.rst`. I leave `VizShield` in a separate file because the very similar and small `VizEvent` is in a separate file. The only difference I see is that `VizShield` is not for users in general. Perhaps they should be put in the same file with context managers?

I don't think we should have a `VizShield` class. The shield feature will only used in `VizTracer` class and it should be so simple with `setignorestackcounter()`. The current structure is over-complicated. `VizEvent` might not be the best design, but the user can directly use it.

I think the shield code can be as simple as

```python
from contextlib import contextmanager
...
@contextmanager
def shield_ignore(self):
    prev_ignore_stack = self.setignorestackcounter(0)
    yield
    self.setignorestackcounter(prev_ignore_stack)
```

I don't think you need to care about whether the tracer is enabled or in log_sparse mode, shield is shield - the caller is responsible for checking the environment. Also you imported `VizShield` in a file that does not use it which caused a lint error. Should be trivial to fix.

gaogaotiantian:
One final last thing and we are done. Could you move `from contextlib import contextmanager` before `from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Union`? Or, feel free to do `import contextlib` and use `contextlib.contextmanager` if that's what you prefer.

We are grouping `import XX` and `from XX import YY` separately which is what stdlib does (mostly).

AlexeyKozhevin:
I conducted more detailed testing and ran local tests. Unfortunately, I have to return to `_VizShield` decorator due to additional redundant calls. 

At the last commit implementation we need to compensate stack of calls by adding or subtracting of 2 in `setignorestackcounter` and by adding 1 in the `__enter__` of `_VizShield` (see comments in the code)

In the case of using `@contextlib` we have to work with additional generator and `next` calls (see image below). The generator makes the logic of calls compensation not so straightforward and I haven't figured out how to handle it (just increasing the `setignorestackcounter` parameter doesn't work).

Screenshot 2023-10-31 at 22 39 20

I concluded that syntactic sugar of `contexlib` makes it difficult to understand what `shield_ignore` does and stopped at the current implementation. What do you think about it?

Also wondering what you think about the `_VizShield` place? I haven't found any places where classes are defined locally, but here the class is purely disposable, so I thought it would be appropriate to define it inside `shield_ignore`

gaogaotiantian:
Ah, right. The `with` command is actually a series of calls before/after the statement. In that case, I'd say just get rid of the `with` command. The compensation is easier to understand with the class implementation, but still not easy enough - it's fragile.

Let's do
```python
def shield_ignore(self, func, *args, **kwargs):
    prev_counter = self.setignorestackcounter(0)
    func(*args, **kwargs)
    self.setignorestackcounter(prev_counter)
```

Could you write a test to make sure `setignorestackcounter` is not logged in the trace? Basically a test that would fail if you did not do the compensation in `setignorestackcounter` correctly - that would be helpful in the future to prevent regression.

AlexeyKozhevin:
Do you mean
```
    def shield_ignore(self, func, *args, **kwargs):
        prev_ignore_stack = self.setignorestackcounter(0)
        res = func(*args, **kwargs)
        self.setignorestackcounter(prev_ignore_stack)
        return res
```
and
```
    return shield_ignore(func, *args, **kwargs))
```
instead of 
```
with local_tracer.shield_ignore():
    return func(*args, **kwargs)
```
?

And yes, I will add a test.

gaogaotiantian:
Yes that's what I meant - also makes to code simpler I believe.

AlexeyKozhevin:
I have extended the tests and also modified the existing `test_basic`. Since the name of a function is not just `f`, but also the path to the module where it is defined, the check in `check_func` always passed. And it seems that in general the wrong things were checked there :)

gaogaotiantian:
Okay now we only have a couple of lint issues - you can install `flake8` and `mypy` in your environment and do `make lint` to check it quickly. For the `cellmagic.py`, I don't know what happened there, feel free to figure it out, but if it's too complicated, just do an inline ignore (I think it's `# type: ignore`? There are a couple in the source files already).

For the decorator, I think `_log_sparse_wrapper` is guaranteed a not `None` `func` right? So removing the `Optional` for the `func` argument typing might fix most of the problems.

If you are not familiar with typing, you can send the typing you can't solve here and I'll take a look. Thanks!

AlexeyKozhevin:
`flake8` and `mypy` say it's okay now.

Unfortunately, I have no idea what the problem with `cellmagic.py` so I have to set `# type: ignore`.

gaogaotiantian:
Thank you for your contribution, it's been a long journey and I'm really glad that log_sparse now works with the inline tracer and jupyter!

AlexeyKozhevin:
Thank you for advices and review! It was really interesting.
We already use viztracer in our ML projects in notebook. Some things have become much easier to profile

PR #384, opened 2023-12-14
Skip invalid json files in ReportBuilder

TTianshun:
1. Now the `ReporterBuilder` will ignore the corrupted json and only combine those valid json. It will warn user if there's corrupted json file.
image

2. When waiting for child, the main process will find all subprocess recursively and check if the subproceess has terminated.

### DIFF
--- a/src/viztracer/report_builder.py
+++ b/src/viztracer/report_builder.py
@@ -84,10 +84,11 @@ class ReportBuilder:
             same_line_print("Combining trace data")
         if self.combined_json:
             return
-        if not self.jsons and not self.invalid_json_paths:
-            raise ValueError("Can't get report of nothing")
-        elif not self.jsons:
-            raise ValueError("No valid json files found")
+        if not self.jsons:
+            if self.invalid_json_paths:
+                raise ValueError("No valid json files found")
+            else:
+                raise ValueError("Can't get report of nothing")
         if self.align:
             for one in self.jsons:
                 self.align_events(one["traceEvents"])
--- a/src/viztracer/report_builder.py
+++ b/src/viztracer/report_builder.py
@@ -77,7 +77,7 @@ class ReportBuilder:
                     except json.JSONDecodeError:
                         self.invalid_json_paths.append(j)
                 if len(self.invalid_json_paths) > 0:
-                    self.final_messages.append(("corrupted_json", {"paths": self.invalid_json_paths}))
+                    self.final_messages.append(("invalid_json", {"paths": self.invalid_json_paths}))
 
     def combine_json(self) -> None:
         if self.verbose > 0:
@@ -87,9 +87,7 @@ class ReportBuilder:
         if not self.jsons and not self.invalid_json_paths:
             raise ValueError("Can't get report of nothing")
         elif not self.jsons:
-            # print corrupted json files
-            self.print_messages()
-            raise ValueError("All json files are corrupted")
+            raise ValueError("No valid json files found")
         if self.align:
             for one in self.jsons:
                 self.align_events(one["traceEvents"])
@@ -224,10 +222,10 @@ class ReportBuilder:
                         color_print("OKGREEN", f"vizviewer \"{report_abspath}\"")
                     else:
                         color_print("OKGREEN", f"vizviewer {report_abspath}")
-                elif msg_type == "corrupted_json":
+                elif msg_type == "invalid_json":
                     print("")
-                    color_print("WARNING", "Found and ignored corrupted json file, you may lost some process data.")
-                    color_print("WARNING", "Corrupted json file:")
+                    color_print("WARNING", "Found and ignored invalid json file, you may lost some process data.")
+                    color_print("WARNING", "Invalid json file:")
                     for msg in msg_args["paths"]:
                         color_print("WARNING", f"    {msg}")
                     print("")
--- a/src/viztracer/report_builder.py
+++ b/src/viztracer/report_builder.py
@@ -47,6 +47,7 @@ class ReportBuilder:
         self.align = align
         self.minimize_memory = minimize_memory
         self.jsons: List[Dict] = []
+        self.invalid_json_paths: List[str] = []
         self.json_loaded = False
         self.final_messages: List[Tuple[str, Dict]] = []
         if not isinstance(data, (dict, list, tuple)):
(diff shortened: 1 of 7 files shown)


TTianshun:
Seemd that there's not an easy way to get children on mac with pure python. I skipped find all subprocess on mac.

gaogaotiantian:
Okay. I think the code to check existing pids on Windows is a bit too invasive. Bringing up another subprocess to run a cmd based command probably makes this more complicated than it is supposed to be, and the problem it solves is not as important.

We can deal with `SIGTERM` on Linux, so Linux actually has a much smaller issue compared to Windows. The Linux code is cleaner, but less useful.

The benefit to introduce such a large piece of fragile code is not significant enough - we still can't recover the data, we only know the process actually finished. I'm not saying it's not useful, but the code might be hard to maintain in a long term.

The skipping of corrupted json report on the other hand, is much useful - it keeps the data from finished processes. There are some stuff that I'd like to change on the UI side, but overall I think that makes sense.

So for now, could you remove the pid checking part, just leave the report piece? I think Ctrl+C to cancel is enough.

However, there might be a cleaner way to do this, without relying on the cmdlets and Linux path for processes so it's cross-platform. We can check the size of the json report. If it has not changed for a while, then the process must be dead (at least we can warn the users). Or, we can at least list out the processes that we are waiting for (requires some techniques to refresh the line), and the user might realize it's stuck.

For this PR, let's do one thing at a time. Do the json report skipping first, then let's do the feature to detect dead processes.

TTianshun:
> Okay. I think the code to check existing pids on Windows is a bit too invasive. Bringing up another subprocess to run a cmd based command probably makes this more complicated than it is supposed to be, and the problem it solves is not as important.
> 
> We can deal with `SIGTERM` on Linux, so Linux actually has a much smaller issue compared to Windows. The Linux code is cleaner, but less useful.
> 
> The benefit to introduce such a large piece of fragile code is not significant enough - we still can't recover the data, we only know the process actually finished. I'm not saying it's not useful, but the code might be hard to maintain in a long term.
> 
> The skipping of corrupted json report on the other hand, is much useful - it keeps the data from finished processes. There are some stuff that I'd like to change on the UI side, but overall I think that makes sense.
> 
> So for now, could you remove the pid checking part, just leave the report piece? I think Ctrl+C to cancel is enough.
> 
> However, there might be a cleaner way to do this, without relying on the cmdlets and Linux path for processes so it's cross-platform. We can check the size of the json report. If it has not changed for a while, then the process must be dead (at least we can warn the users). Or, we can at least list out the processes that we are waiting for (requires some techniques to refresh the line), and the user might realize it's stuck.
> 
> For this PR, let's do one thing at a time. Do the json report skipping first, then let's do the feature to detect dead processes.

Yes the current code to check existing pids is not suitable. For the children of multiprocessing. Maybe we can use `multiprocessing.active_children()` to get it's process info. For subprocess, there's not a similar method, so I use this to get all the children. Maybe we can add a similar method on subprocess and patch it. If we use file size to check if the process is running, there's no json file if the child has not yet run to it's finalizer.

gaogaotiantian:
Seems like you can use the API interface for WMI https://stackoverflow.com/a/568589. That's a much cleaner way to check the existing PIDs. Combining with the current `pid_exists` for POSIX, I think that's good enough.

So - get the pids to wait for from the `viztmp` file -> that's deterministic and we can control that part, less dependency. For each pid, check whether that pid exists by one function.

(The windows api is not the fastest, you may either need some cache mechanism, or have a new function `pids_exist` to batch process them)

However, do this in a separate PR. We want to keep each PR as single purpose as possible.

src/viztracer/report_builder.py
```diff
@@ -67,10 +67,16 @@ def load_jsons(self) -> None:
                 self.jsons = [get_json(self.data)]
             elif isinstance(self.data, (list, tuple)):
                 self.jsons = []
+                corrupted_jsons = []
```
gaogaotiantian, line 70: Let's call it `self.invalid_json_paths` as this does not store json data and it might not be "corrupted" - could be not json at all.

We need this in the future because if there's no valid json files, we will raise an exception and the final message won't be printed.

tests/test_report_builder.py
```diff
@@ -70,6 +71,16 @@ def test_invalid_json(self):
...
 
+    def test_corrupted_json(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            corrupted_json_path = os.path.join(os.path.dirname(__file__), "data", "fib.py")
+            valid_json_path = os.path.join(os.path.dirname(__file__), "data", "multithread.json")
+            corrupted_json_file = shutil.copy(corrupted_json_path, os.path.join(tmpdir, "corrupted.json"))
+            valid_json_file = shutil.copy(valid_json_path, os.path.join(tmpdir, "valid.json"))
+            rb = ReportBuilder([corrupted_json_file, valid_json_file], verbose=1)
+            with io.StringIO() as s:
+                rb.save(s)
```
gaogaotiantian, line 82: We should check for some printed messages here. This test only assures no exceptions raised.

gaogaotiantian (changes requested):
Like I mentioned, in `combine_json()`, if there's no valid json, but there are inputs (all of them invalid), we should generate a better message to the users.

TTianshun:
> Like I mentioned, in `combine_json()`, if there's no valid json, but there are inputs (all of them invalid), we should generate a better message to the users.

Now it will still raise an exception if all of the json files are invalid. I've changed the error message and  make it print warning before raise an exception. Just like this picture: 
image

src/viztracer/report_builder.py
```diff
@@ -67,18 +68,28 @@ def load_jsons(self) -> None:
...
             same_line_print("Combining trace data")
         if self.combined_json:
             return
-        if not self.jsons:
+        if not self.jsons and not self.invalid_json_paths:
             raise ValueError("Can't get report of nothing")
+        elif not self.jsons:
+            # print corrupted json files
+            self.print_messages()
+            raise ValueError("All json files are corrupted")
```
gaogaotiantian, line 92:
```suggestion
        if not self.jsons:
            if not self.invalid_json_paths:
                raise ValueError("Can't get report of nothing")
            else:
                raise ValueError("No valid json files found")
```

gaogaotiantian: Changing the exception message should be enough - this is a rather rare case, just separate that from "no files".

src/viztracer/report_builder.py
```diff
@@ -67,18 +68,28 @@ def load_jsons(self) -> None:
...
                 for idx, j in enumerate(self.data):
                     if self.verbose > 0:
                         same_line_print(f"Loading trace data from processes {idx}/{len(self.data)}")
-                    self.jsons.append(get_json(j))
+                    try:
+                        self.jsons.append(get_json(j))
+                    except json.JSONDecodeError:
+                        self.invalid_json_paths.append(j)
+                if len(self.invalid_json_paths) > 0:
+                    self.final_messages.append(("corrupted_json", {"paths": self.invalid_json_paths}))
```
gaogaotiantian, line 80: Do a full search on this PR and change all "corrupted" to "invalid".

src/viztracer/report_builder.py
```diff
@@ -67,18 +68,26 @@ def load_jsons(self) -> None:
...
 
     def combine_json(self) -> None:
         if self.verbose > 0:
             same_line_print("Combining trace data")
         if self.combined_json:
             return
-        if not self.jsons:
+        if not self.jsons and not self.invalid_json_paths:
             raise ValueError("Can't get report of nothing")
+        elif not self.jsons:
```
gaogaotiantian, line 89: Why do you hate nested if so much :) . You are evaluating `self.jsons` twice (performance wise it's not a big deal, but logic wise it's not clean).

Just check `self.jsons` first, and raise different errors based on `self.invalid_json_paths`. Nested if works in many cases.

TTianshun: > Why do you hate nested if so much :) . You are evaluating `self.jsons` twice (performance wise it's not a big deal, but logic wise it's not clean).
> 
> Just check `self.jsons` first, and raise different errors based on `self.invalid_json_paths`. Nested if works in many cases.

Sorry, I didn't notice that. And thank you for correcting my bad habit.

gaogaotiantian:
Good job on this, thanks!

PR #388, opened 2023-12-25
check pid exists when waiting for child

TTianshun:
Added child check when waiting for children to finish, in case that child failed to delete viztmp file. The current `pid_exists` has a issue that if a process return 259 (STILL_ACTIVE) as exit code (which is a really rare case that is nearly impossible I guess).

### DIFF
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -398,7 +398,7 @@ if __name__ == '__main__':
     p.start()
     # The main process will join the child in multiprocessing.process._children.
     # This is a hack to make sure the main process won't join the child process,
-    # so we can test the wait_children_finish function
+    # so we can test the VizUI.wait_children_finish function
     multiprocessing.process._children = set()
     time.sleep(1)
 """
@@ -418,7 +418,7 @@ if __name__ == '__main__':
     p.start()
     # The main process will join the child in multiprocessing.process._children.
     # This is a hack to make sure the main process won't join the child process,
-    # so we can test the wait_children_finish function
+    # so we can test the VizUI.wait_children_finish function
     multiprocessing.process._children = set()
     time.sleep(1)
 """
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -396,6 +396,9 @@ def target():
 if __name__ == '__main__':
     p = multiprocessing.Process(target=target)
     p.start()
+    # The main process will join the child in multiprocessing.process._children.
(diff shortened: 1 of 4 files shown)


gaogaotiantian (changes requested):
Overall this looks good, a couple of comments.

tests/test_regression.py
```diff
@@ -384,3 +384,53 @@ def test_escape_string(self):
...
+def target():
+    time.sleep(3)
+    os.kill(os.getpid(), signal.SIGTERM)
+
+if __name__ == '__main__':
+    p = multiprocessing.Process(target=target)
+    p.start()
+    # The main process will join the child in multiprocessing.process._children.
+    # This is a hack to make sure the main process won't join the child process,
+    # so we can test the wait_children_finish function
```
gaogaotiantian, line 421: For both comments, let's mention the class (I think it's `VizUI.wait_children_finish`?). Because we are discussing in the context of `multiprocessing` here and it's easy for people to mistaken this function as a `multiprocessing` function.

gaogaotiantian (changes requested):

TTianshun:
Fixed it!

gaogaotiantian:
I believe this PR is making the nightly validation unstable - https://github.com/gaogaotiantian/viztracer/actions/runs/7539946132/job/20523513962 (not sure if you have access to it).

The issue is the child process only lives for 1 second, and the parent process might need more to reach the trace collection part. github actions is normally much slower than our PC. There are multiple tests that use different timeout on github actions (search for `GITHUB_ACTIONS`) and we should probably do the same here. (To not increase the local full-test time and stablize the remote one).

TTianshun:
> I believe this PR is making the nightly validation unstable - https://github.com/gaogaotiantian/viztracer/actions/runs/7539946132/job/20523513962 (not sure if you have access to it).
> 
> The issue is the child process only lives for 1 second, and the parent process might need more to reach the trace collection part. github actions is normally much slower than our PC. There are multiple tests that use different timeout on github actions (search for `GITHUB_ACTIONS`) and we should probably do the same here. (To not increase the local full-test time and stablize the remote one).

I agree with that if using a much slower PC the test will fail. It takes about 0.4 seconds on my computer between `_winapi.CreateProcess` and the child starting. The time may be longer on github actions. So when the parent collects the data, the child might haven't created viztmp file.

PR #391, opened 2024-01-04
make subprocess child shows different process name

TTianshun:
The previous name for cubprocess child is MainProcess, which make it not easy to distinguish MainProcess and subprocess child.
image

This PR changes the name of subprocess child.
image

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -653,7 +653,7 @@ snaptrace_load(TracerObject* self, PyObject* args)
         PyObject* process_name = NULL;
 
         if (self->process_name) {
-            process_name = PyUnicode_FromString(self->process_name);
+            process_name = self->process_name;
         } else {
             PyObject* current_process_method = PyObject_GetAttrString(multiprocessing_module, "current_process");
             if (!current_process_method) {
@@ -906,7 +906,7 @@ snaptrace_dump(TracerObject* self, PyObject* args)
     {
         PyObject* process_name = NULL;
         if (self->process_name) {
-            process_name = PyUnicode_FromString(self->process_name);
+            process_name = self->process_name;
         } else {
             PyObject* current_process_method = PyObject_GetAttrString(multiprocessing_module, "current_process");
             if (!current_process_method) {
@@ -1158,7 +1158,7 @@ snaptrace_config(TracerObject* self, PyObject* args, PyObject* kw)
     int kw_verbose = -1;
     int kw_max_stack_depth = 0;
     char* kw_lib_file_path = NULL;
-    char* kw_process_name = NULL;
+    PyObject* kw_process_name = NULL;
     PyObject* kw_include_files = NULL;
     PyObject* kw_exclude_files = NULL;
     int kw_ignore_c_function = -1;
@@ -1168,7 +1168,7 @@ snaptrace_config(TracerObject* self, PyObject* args, PyObject* kw)
     int kw_log_async = -1;
     int kw_trace_self = -1;
     double kw_min_duration = 0;
-    if (!PyArg_ParseTupleAndKeywords(args, kw, "|isiOOppppppds", kwlist,
+    if (!PyArg_ParseTupleAndKeywords(args, kw, "|isiOOppppppdO", kwlist,
             &kw_verbose,
             &kw_lib_file_path,
             &kw_max_stack_depth,
@@ -1205,20 +1205,12 @@ snaptrace_config(TracerObject* self, PyObject* args, PyObject* kw)
         strcpy(self->lib_file_path, kw_lib_file_path);
     }
 
-    if (kw_process_name) {
+    if (kw_process_name && kw_process_name != Py_None) {
         if (self->process_name) {
-            PyMem_FREE(self->process_name);
+            Py_DECREF(self->process_name);
         }
-        if (strlen(kw_process_name) != 0) {
-            self->process_name = PyMem_Calloc((strlen(kw_process_name) + 1), sizeof(char));
-            if (!self->process_name) {
-                printf("Out of memory!\n");
-                exit(1);
-            }
-            strcpy(self->process_name, kw_process_name);
-        } else {
-            self->process_name = NULL;
-        }
+        self->process_name = kw_process_name;
+        Py_INCREF(self->process_name);
     }
 
     if (kw_ignore_c_function == 1) {
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -650,20 +650,26 @@ snaptrace_load(TracerObject* self, PyObject* args)
         PyObject* dict = PyDict_New();
         PyObject* args = PyDict_New();
         PyObject* process_name_string = PyUnicode_FromString("process_name");
-        PyObject* current_process_method = PyObject_GetAttrString(multiprocessing_module, "current_process");
-        if (!current_process_method) {
-            perror("Failed to access multiprocessing.current_process()");
-            exit(-1);
-        }
-        PyObject* current_process = PyObject_CallObject(current_process_method, NULL);
-        if (!current_process_method) {
-            perror("Failed to access multiprocessing.current_process()");
-            exit(-1);
-        }
-        PyObject* process_name = PyObject_GetAttrString(current_process, "name");
+        PyObject* process_name = NULL;
 
-        Py_DECREF(current_process_method);
-        Py_DECREF(current_process);
+        if (self->process_name){
+            process_name = PyUnicode_FromString(self->process_name);
+        }else{
+            PyObject* current_process_method = PyObject_GetAttrString(multiprocessing_module, "current_process");
+            if (!current_process_method) {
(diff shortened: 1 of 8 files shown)


gaogaotiantian:
`sys.argv[0]` might be a better candidate here.

gaogaotiantian:
I'm having second thought about this. This is a quick and useful change, but it also has a small side effect - other people's code might rely on `multiprocessing.current_process().name` and we should avoid making changes to stdlib as much as possible.

However, I agree having a different name is helpful, so let's add a variable like `process_name` in `VizTracer` or the c module and access that first when dumping json reports. If the variable is `None`, fall back to current solution. This way we have the least interruption to the client code.

TTianshun:
> I'm having second thought about this. This is a quick and useful change, but it also has a small side effect - other people's code might rely on `multiprocessing.current_process().name` and we should avoid making changes to stdlib as much as possible.
> 
> However, I agree having a different name is helpful, so let's add a variable like `process_name` in `VizTracer` or the c module and access that first when dumping json reports. If the variable is `None`, fall back to current solution. This way we have the least interruption to the client code.

OK, I can try it in the c module.

gaogaotiantian:
The C module part is fine, but do not pass the process name as a cmdline argument, that's totally unnecessary. Just use your original method to set it in `main()` if it's a subprocess. cmdline arguments are very fragile.

Also, check the coding styles in the C code, especially spaces around `else`.

gaogaotiantian:
Let's remove the optional command line argument to set process name, that seems a bit too much for me. In a real subprocess case, there's no way for the user to properly set it. If we don't need it internally, we don't need it.

Also, it seems like you are converting the string back and forth, from Python string to C string then back. Why not simply keep it as a Python string? You can parse that as an object and you don't need to free/reallocate memory anymore, just reference change. By parsing it as an object, you can also use `None` as the default value for "use other values", instead of `""`, which is a bit like a real value.

TTianshun:
> Let's remove the optional command line argument to set process name, that seems a bit too much for me. In a real subprocess case, there's no way for the user to properly set it. If we don't need it internally, we don't need it.
> 
> Also, it seems like you are converting the string back and forth, from Python string to C string then back. Why not simply keep it as a Python string? You can parse that as an object and you don't need to free/reallocate memory anymore, just reference change. By parsing it as an object, you can also use `None` as the default value for "use other values", instead of `""`, which is a bit like a real value.

Finished it. The commandline way is indeed kind of fragile.

src/viztracer/modules/snaptrace.c
```diff
@@ -1190,6 +1205,14 @@ snaptrace_config(TracerObject* self, PyObject* args, PyObject* kw)
         strcpy(self->lib_file_path, kw_lib_file_path);
     }
 
+    if (kw_process_name && kw_process_name != Py_None) {
+        if (self->process_name) {
+            Py_DECREF(self->process_name);
+        }
```
gaogaotiantian, line 1211:
```suggestion
        Py_XDECREF(self->process_name);
```

src/viztracer/modules/snaptrace.c
```diff
@@ -1190,6 +1205,14 @@ snaptrace_config(TracerObject* self, PyObject* args, PyObject* kw)
...
+        self->process_name = kw_process_name;
+        Py_INCREF(self->process_name);
```
gaogaotiantian, line 1213:
```suggestion
        Py_INCREF(kw_process_name);
        Py_XSETREF(self->process_name, kw_process_name);  
```

src/viztracer/viztracer.py
```diff
@@ -43,7 +43,8 @@ def __init__(self,
                  dump_raw: bool = False,
                  sanitize_function_name: bool = False,
                  output_file: str = "result.json",
-                 plugins: Sequence[Union[VizPluginBase, str]] = []) -> None:
+                 plugins: Sequence[Union[VizPluginBase, str]] = [],
+                 process_name: Optional[str] = None) -> None:
```
gaogaotiantian, line 47: Put `process_name` before `output_file` -> there's no strict order here, but I always feel like `output_file` and `plugins` should be at the end of the list.

src/viztracer/modules/snaptrace.c
```diff
@@ -898,22 +904,29 @@ snaptrace_dump(TracerObject* self, PyObject* args)
...
-            perror("Failed to access multiprocessing.current_process()");
-            exit(-1);
-        }
-        PyObject* current_process = PyObject_CallObject(current_process_method, NULL);
-        if (!current_process_method) {
-            perror("Failed to access multiprocessing.current_process()");
-            exit(-1);
+        PyObject* process_name = NULL;
+        if (self->process_name) {
+            process_name = self->process_name;
```
gaogaotiantian, line 909: You are losing reference here. You need to increment the reference because you are decreasing it later.

src/viztracer/modules/snaptrace.c
```diff
@@ -650,20 +650,26 @@ snaptrace_load(TracerObject* self, PyObject* args)
...
-            perror("Failed to access multiprocessing.current_process()");
-            exit(-1);
-        }
-        PyObject* process_name = PyObject_GetAttrString(current_process, "name");
+        PyObject* process_name = NULL;
 
-        Py_DECREF(current_process_method);
-        Py_DECREF(current_process);
+        if (self->process_name) {
+            process_name = self->process_name;
```
gaogaotiantian, line 656: New reference here as well

src/viztracer/modules/snaptrace.c
gaogaotiantian: Maybe also add a unicode check here - to make sure toe process name given is a string, otherwise raise an exception.

tests/test_multiprocess.py
```diff
@@ -240,13 +246,26 @@ def test_child_process(self):
...
     def test_module(self):
         self.template(["viztracer", "-o", "result.json", "cmdline_test.py"],
                       expected_output_file="result.json",
                       expected_stdout=".*100 loops.*",
                       script=file_subprocess_module)
 
+    def test_code(self):
+        self.template(["viztracer", "-o", "result.json", "cmdline_test.py"],
+                      expected_output_file="result.json",
+                      script=file_subprocess_code)
```
gaogaotiantian, line 267: Let's also check the process name of the subprocess on this, as it's known - should be `python` I believe?

TTianshun: Yes, the current process_name is `site-packages/viztracer/__main__.py` because `sys.argv` is processed in `VizUI.run()`, which is before the config.

gaogaotiantian: Having a process with a name `python` is almost as bad as `MainProcess`. Let's set the `process_name` in `run_code`. Update `self.init_kwargs` before create the `VizTracer` instance. The subprocess would be distinguishable with the distinct names.

TTianshun: If setting `process_name` in `run_code`. For module, the `process_name` is module name. For run python file, the `process_name` is python file name. For run code by `-c`, the process_name is `-c`.

gaogaotiantian: Yeah `"-c"` is a bit weird. Let's do another check for `cmd_string` and if it's set, set the process name to `"python -c"`.

gaogaotiantian:
The `python -c` part is not covered, probably due to the multiprocessing handling mechanism of coveragepy. Could you solve that? If it's too much for you, I can take a look at it.

TTianshun:
> The `python -c` part is not covered, probably due to the multiprocessing handling mechanism of coveragepy. Could you solve that? If it's too much for you, I can take a look at it.

Seemed that coverage didn't support subprocess as concurrency libraries. Maybe we should patch the command line for subprocess in coverage run.

gaogaotiantian:
A couple of suggestions with the tests and we can be done with this.

1. For `test_child_process`, instead of checking what the process name is *not*, check what it *is*. It should be `child.py` right?
2. Use `check_func` argument for `template` for `test_module`. Also rename this test to `test_module_process_name`.
3. `test_code` is also too general, at least it should be `test_code_string`. It's okay at this point to use `--subprocess_child` for the coverage, but we should still keep the normal test that actually calls `subprocess` as function test. They can live in the same test method because they share the same check function. For the real test, you can also use `check_func` argument - it's designed for trace data check.

TTianshun:
> A couple of suggestions with the tests and we can be done with this.
> 
> 1. For `test_child_process`, instead of checking what the process name is *not*, check what it *is*. It should be `child.py` right?
> 2. Use `check_func` argument for `template` for `test_module`. Also rename this test to `test_module_process_name`.
> 3. `test_code` is also too general, at least it should be `test_code_string`. It's okay at this point to use `--subprocess_child` for the coverage, but we should still keep the normal test that actually calls `subprocess` as function test. They can live in the same test method because they share the same check function. For the real test, you can also use `check_func` argument - it's designed for trace data check.

For test with `--subprocess_child`, the output file is `result_{pid}.json` and unable to use `check_func`. For `test_module`, I Retain the test as `test_module`. Because the main effect of the test is to test subprocess module.

gaogaotiantian:
Great work, thanks for the contribution.

PR #397, opened 2024-01-13
Support building docs in pdf

F-park:
user can download docs in pdf in readthedocs

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -19,6 +19,10 @@ sphinx:
 #mkdocs:
 #  configuration: mkdocs.yml
 
+# Optionally build your docs in additional formats such as PDF and ePub
+formats:
+  - pdf
+
 # Optionally set the version of Python and requirements required to build your docs
 python:
   install:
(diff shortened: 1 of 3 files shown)


gaogaotiantian:
This seems like a very random feature with no request. I don't think anyone is using pdf docs.

F-park:
> This seems like a very random feature with no request. I don't think anyone is using pdf docs.

I need the pdf docs and I revert the unnecessary modify.

gaogaotiantian:
This seems like a very odd request, but the code change is small. What's the result of this? Will it just work so you can download pdf from the readthedocs website?

F-park:
It just a [build config](https://docs.readthedocs.io/en/stable/config-file/v2.html#formats), and I download pdf from the readthedocs website.

image

.readthedocs.yml
```diff
@@ -19,6 +19,10 @@ sphinx:
 #mkdocs:
 #  configuration: mkdocs.yml
 
+# Optionally build your docs in additional formats such as PDF and ePub
```
gaogaotiantian, line 22:
```suggestion
# Enable pdf download on readthedocs
```

PR #398, opened 2024-01-13
Add paths to make CI strict

F-park:
add `on.(push|pull_request).paths`


.github/workflows/lint.yml
```diff
@@ -3,7 +3,19 @@ name: lint
 on:
   push:
     branches: [ master ]
+    paths:
```
gaogaotiantian, line 6: lint is fast enough that this path restriction is not necessary. It's too wordy and the benefit is just not enough.

.github/workflows/documentation-links.yml
```diff
@@ -0,0 +1,21 @@
...
+permissions:
+  pull-requests: write
+
+jobs:
+  documentation-links:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: readthedocs/actions/preview@v1
+        with:
+          project-slug: "viztracer-preview"
```
gaogaotiantian, line 20: I don't believe this works without the actual readthedocs project.

.github/workflows/docs.yml
```diff
@@ -3,7 +3,9 @@ name: docs
 on:
   push:
     branches: [ master ]
-  pull_request:
+    paths:
```
gaogaotiantian, line 6: Leave the `pull_request` as it is because the readthedocs project is not set up for build & preview. Try to finish one thing for one PR.

gaogaotiantian (changes requested):

.github/workflows/documentation-links.yml
F-park: I cannot understand, it can work in other [project](https://github.com/dranjan/python-plyfile/blob/master/.github/workflows/pull-request-doc-preview.yml)

gaogaotiantian: Yeah, because https://python-plyfile.readthedocs.io/ exists - you need to register the project on readthedocs first to make this work. There's no `viztracer-preview` project on readthedocs so it won't work. Try it locally and see.

F-park: You are right, I should learn how it works and I remove the file.

PR #400, opened 2024-01-20
Increace sleep time of TestWaitForChild on github actions

TTianshun:


gaogaotiantian:
I'd prefer to check the env var inside the code string. It would be clear to see what the logic is, instead of having to check the caller to determine the sleep time.

TTianshun:
Seemed this fail is not related with this PR

PR #432, opened 2024-05-09
Add support for building on FreeBSD

zebrapurring:
This PR adds support for building the package on FreeBSD through [`PTHREAD_NP`](https://man.freebsd.org/cgi/man.cgi?query=pthread_np&sektion=3&manpath=FreeBSD+13.2-RELEASE+and+Ports/1000), since it is not compatible with the `gettid` syscall.

Tested on FreeBSD 13.

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -8,8 +8,10 @@
 #include <time.h>
 #if _WIN32
 #include <windows.h>
-#elif __APPLE
+#elif defined(__APPLE__)
 #include <pthread.h>
+#elif defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


gaogaotiantian:
Thanks. We don't have official support for FreeBSD because there's no github action runner for it. However, the patch is easy to follow and does not affect the platforms we do support. I'll merge it for people who use FreeBSD :)

zebrapurring:
Thanks! Some features like attaching to processes are still not working on FreeBSD, but at least this should allow installing the package via Pip and generating traces.

gaogaotiantian:
Attaching would be hard. That is some nasty hack feature anyway.

Issue #441, opened 2024-06-07
[optimize] always add quotes when printing file path


FunnySaltyFish:
I'm a new user for `viztracer`, when using the project, I found a small improvement.

Currently, the report builder will add quotes when there is space in the file path, as shown in the code below, due to https://github.com/gaogaotiantian/viztracer/issues/324:

```python
# src/viztracer/report_builder.py
if " " in report_abspath:
    color_print("OKGREEN", "vizviewer \"{}\"".format(report_abspath))
else:
    color_print("OKGREEN", "vizviewer {}".format(report_abspath))
```

However, when I using the `Git Bash` on `Windows`, the separator will be `\`, which will be considered as escape symbol in git bash. The command like: `vizviewer D:\python\result.json` will throw `File D:pythonresult.json does not exist!` error. By simply adding the quotes, we can solve the problem: `vizviewer "D:\python\result.json"`. This new command is valid for `git bash`, `powershell` and `cmd`.

So i'm wondering whether it's better to add quotes for all output paths, no matter there contains spaces or not. Or if there are some reasons not to do so. I'm glad to give a PR for it. Thanks.

gaogaotiantian:
Quotes look bad... I'd prefer not to always add quotes around the path. Git bash is a relatively rare environment - it uses bash but also windows path? That fact by itself is a bit complicated.

FunnySaltyFish:
Referring to what discussed in #442, I'll also close this issue. Thanks for replying.

PR #442, opened 2024-06-09
Always use quotes around file paths in viewer_main function

ZhouQiang19980220:
Regardless of whether the absolute path contains spaces, it should always be used with quotes. This ensures consistent behavior across different environments and avoids potential issues with path parsing in shells and command line interfaces.

- Updated viewer_main function to always wrap the file path in quotes.
- This change improves compatibility and prevents errors related to space and special characters in file paths.

### DIFF
--- a/src/viztracer/report_builder.py
+++ b/src/viztracer/report_builder.py
@@ -215,10 +215,12 @@ class ReportBuilder:
                 elif msg_type == "view_command":
                     report_abspath = os.path.abspath(msg_args["output_file"])
                     print("Use the following command to open the report:")
-                    if " " in report_abspath:
-                        color_print("OKGREEN", f"vizviewer \"{report_abspath}\"")
-                    else:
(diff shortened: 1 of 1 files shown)


gaogaotiantian:
Like I said in https://github.com/gaogaotiantian/viztracer/issues/441#issuecomment-2156314858, I don't believe this is the way to go. Also double quote does not solve everything. For example, if you have a file with a name `s\\s`, `ls "s\\s"` would be wrong. The correct way is to use single quotes. However, `cmd` does not take single quotes. It's a complicated issue. Not easy to make it always correct on on platforms.

ZhouQiang19980220:
Thank you for your detailed explanation, and I completely agree with your points. 
Ensuring compatibility between Unix-like systems and Windows, especially with special characters in file names, indeed presents a significant challenge.
I appreciate the clarity you’ve provided on this matter. Looking forward to working together to find the best solution.

PR #452, opened 2024-07-21
change README.md

user-11150:


user-11150:
Why to close my PR?

gaogaotiantian:
When you open a PR, you need to be explicit about why you made the change. Your change is "cosmetic" change, which is just a preference. If you want to be more technical, Linux and Windows are always used capitalized, there's no reason to change them to lower. macOS is now the official representation by Apple, and it used to be Mac OS. I don't like the format and it does not hurt the readability - it causes no confusion in any case.

I don't believe this is a PR that I want to merge, and you did not comment on why you wanted to do this - that's why I just closed the PR.

user-11150:
> When you open a PR, you need to be explicit about why you made the change. Your change is a "cosmetic" change, which is just a preference. If you want to be more technical, Linux and Windows are always used capitalized, there's no reason to change them to lower. macOS is now the official representation by Apple, and it used to be Mac OS. I don't like the format and it does not hurt the readability - it causes no confusion in any case.
> 
> I don't believe this is a PR that I want to merge, and you did not comment on why you wanted to do this - that's why I just closed the PR.

Ok, I see. I will pay attention to this in the future.

PR #454, opened 2024-07-24
Update github actions

yihong0618:
This patch fix the year change on README
fix lint error on mypy check, also fix the warning on GitHub Actions


gaogaotiantian:
I've been so lazy to fix this. The lint is not working properly anyway. I'll fix this next in the next commit (probably fixing the lint error). Thanks anyway!

yihong0618:
I would like to fix the mypy error in the next commit

PR #465, opened 2024-08-10
Add special python args to subprocess patch

nielstron:
Awesome project! I was recently adapting the patch utilities for subprocesses for a personal project and noticed that in some particular cases (i.e. when tracing `tox`) the patches don't work. The issue was resolved when I added the following python commands that take parameters (`-X`, `-W` and `--check-hash-based-pycs`) as special cases - otherwise their argument would be parsed as "script" parameter.

**Example**  
`python3 -X dev -m pytest` would be split to `python3 -X (patch-script) dev -m pytest` instead of the correct `python3 -X dev (patch-script) -m pytest`.

### DIFF
--- a/src/viztracer/patch.py
+++ b/src/viztracer/patch.py
@@ -55,6 +55,8 @@ def patch_subprocess(viz_args: list[str]) -> None:
 
             # -pyopts
             py_args.append(arg)
+            if arg in ("-X", "-W", "--check-hash-based-pycs"):
+                py_args.append(next(args_iter, None))
 
         if script:
             return [sys.executable, *py_args, "-m", "viztracer", "--quiet", *viz_args, "--", script, *args_iter]


gaogaotiantian:
I don't have my dev machine near me recently so I'll have to test this myself later. However, I don't believe the solution is the right way to go. I believe the current regex is meant to detect the actual script part - so it should detect the `-m pytest` part and insert all the other py_args before `-m`. If it's not, then we should fix the regex, instead of adding an exception for a specific option. I believe it's because the file regex is too loose on detecting files.

nielstron:
I see and I tried this before, however I think it's not possible using the current approach. For example a valid parameterization for python is `python3 -X dev`.

The current approach splits the arguments by space and thus encounters first `-X` then `dev`. But it is perfectly possible that there is a script in the current directory with the filename `dev`. I believe it's not possible to distinguish the two situations looking only at the current split. But feel free to try and implement it the other way around.

gaogaotiantian:
Okay that makes sense. But in that case you probably need to take out all the dash options with arguments before identifying the file. I can work on this later when I finish my moving.

gaogaotiantian:
Hi, sorry it took me so long to get back at this. I think the code should work fine. Do you mind adding a regression test for this?

nielstron:
I'll have a look. Any suggestions on a related test case that I could adapt?

gaogaotiantian:
It depends on how difficult it is to do a minimum repro. If you can repro this with a normal script on command line usage, checkout tests in `test_cmdline.py`, you can put the test in that file or `test_regression.py`. If it's hard to do it normally, check out `test_patch.py`, there are some coverage tests in that file to test corner cases. Or maybe you can do both :). I think in `test_patch.py` there are tests specifically for parsing subprocess args.

PR #475, opened 2024-10-08
Add support for 3.13t (free threading)

colesbury:
Also drop support for the nogil 3.9 fork.

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -66,15 +66,15 @@ static PyObject* curr_task_getters[2] = {0};
 LARGE_INTEGER qpc_freq;
 #endif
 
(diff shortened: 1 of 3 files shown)


gaogaotiantian:
I'll trust you on this. I'm working on enabling 3.13 CI on viztracer. After that, I'll see if github action supports 3.13t and if the test can actually pass.

colesbury:
`actions/setup-python` unfortunately still doesn't support 3.13t: https://github.com/actions/setup-python/issues/771, but `deadsnakes/action` does.

gaogaotiantian:
Seems like an open discussion. It would be nice if `actions/setup-python` can support it directly, however, it's not the end of the world if I had to use another action to set up 3.13t.

PR #530, opened 2024-12-05
Fix overflow in quicktime calibration loop

### LINKED ISSUE #529: Torch events have wrong time offset on certain machines
Hey!
I've tried recording Torch events with `log_torch`, but for some machines the Torch events are wildly offset from the VizTracer events, often by a factor of 300 to 5,000 hours or so.

My hunch is the uptime of the machine may have a factor (indirectly. My primary suspect is the monotonic clock), since the below trace was obtained from a machine with ~240 days of uptime. I had another trace on a machine with ~190 days of uptime which had a less-prominent shift

VizTracer: 1.0.0
Python: 3.10.4
PyTorch: 2.5.1

image
image
image

Technohacker:
Fixes #529

### DIFF
--- a/src/viztracer/modules/quicktime.c
+++ b/src/viztracer/modules/quicktime.c
@@ -142,12 +142,24 @@ quicktime_init()
         start_ns[i] = get_system_ns();
         int64_t start_after = get_system_ts();
         start_ts[i] = start_before + (start_after - start_before) / 2;
-        t0_ts += start_ts[i];
-        t0_ns += start_ns[i];
     }
 
-    t0_ts /= CALIBRATE_SIZE;
-    t0_ns /= CALIBRATE_SIZE;
+    // Do the expensive average calculation outside the measurement loop
+    int64_t ts_remainder = 0;
+    int64_t ns_remainder = 0;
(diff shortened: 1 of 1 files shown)


src/viztracer/modules/quicktime.c
```diff
@@ -142,12 +142,23 @@ quicktime_init()
...
         int64_t start_after = get_system_ts();
         start_ts[i] = start_before + (start_after - start_before) / 2;
-        t0_ts += start_ts[i];
-        t0_ns += start_ns[i];
     }
 
-    t0_ts /= CALIBRATE_SIZE;
-    t0_ns /= CALIBRATE_SIZE;
+    // Do the expensive average calculation outside the measurement loop
+    int64_t ts_remainder = 0, ns_remainder = 0;
```
gaogaotiantian, line 148: Looks good. Could you split this into two declarations? I'm always trying to avoid declaration like this - it's often okay for a few non-pointer variables, but it gets ugly when it's a pointer or the variable name is too long or there are too many of them. Just as a habit.

Technohacker: Sure! Pushed it now

gaogaotiantian:
Thank you for your contribution!

Issue #534, opened 2025-01-09
Add a sync marker


sergey-miryanov:
Hey, 

First, thanks so much for the awesome tool. It's super helpful and motivating.

I'm using it to compare how well some of our code works on different versions of Python, and it's been working great. I use the align_combine mode to compare results from different versions, and it's been working as expected. But (as you know) py311 is working hard to reduce startup and import times. So, I'm getting shifted results from py310 and py311, and which makes it tricky to compare them. I can avoid it by using the inline approach, but it means using the viztracer explicitly in the code.

What do you think about adding a sync marker and using it as a base for aligning?

If it's doable, I can try to implement this.

Thanks, Sergey

gaogaotiantian:
Yeah, interesting idea. It would be helpful if a sync marker could be added. So my thought is, do not add any trace events, but add a timestamp to `viztracer_metadata` in the final json report. You can add a C API like `set_sync_marker` to the C tracer to record the timestamp and dump it while it dumps trace. In the `align_events` function in `report_builder.py`, add an optional sync timestamp (if not set, the default would be the min of events, which means you can set when is timestamp 0). You can align events with sync marker in this way.

sergey-miryanov:
Thank you for directions, got it.

PR #542, opened 2025-01-25
Implement a sync marker

sergey-miryanov:
Should implement #534

### DIFF
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -392,7 +392,7 @@ class TestCommandLineBasic(CmdlineTmpl):
             aligned_diff = abs(funcs[1]['ts'] - funcs[0]['ts'])
             self.assertLessEqual(aligned_diff, 1000.0, str(data))
 
-        def test_align(extra_args):
+        for extra_args in [[], ["--dump_raw"]]:
             with tempfile.TemporaryDirectory() as tmpdir:
                 res1_filename = os.path.join(tmpdir, 'res1.json')
                 res2_filename = os.path.join(tmpdir, 'res2.json')
@@ -414,12 +414,8 @@ class TestCommandLineBasic(CmdlineTmpl):
                 self.template(
                     [sys.executable, "-m", "viztracer", "--align_combine", res1_filename, res2_filename],
                     expected_output_file="result.json",
-                    script='',
                     check_func=expect_aligned_to_sync_marker,
                 )
-
-        test_align([])
-        test_align(['--dump_raw'])
 
     def test_tracer_entries(self):
         self.template([sys.executable, "-m", "viztracer", "--tracer_entries", "1000", "cmdline_test.py"])
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -368,7 +368,6 @@ class TestCommandLineBasic(CmdlineTmpl):
         """)
 
         def expect_sync_marker(data):
-            self.assertIsNotNone(data['viztracer_metadata'].get('sync_marker'))
             self.assertGreater(data['viztracer_metadata'].get('sync_marker'), 0)
 
         self.template(
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -350,22 +350,22 @@ class TestCommandLineBasic(CmdlineTmpl):
                       expected_output_file="result.json")
 
     def test_set_sync_marker(self):
-        test_script = """
-from viztracer import get_tracer
+        test_script = textwrap.dedent("""
+            from viztracer import get_tracer
 
-if get_tracer() is not None:
-    get_tracer().set_sync_marker()
+            if get_tracer() is not None:
+                get_tracer().set_sync_marker()
 
-def fib(n):
-    if n < 2:
-        return 1
-    return fib(n-1) + fib(n-2)
-fib(5)
+            def fib(n):
+                if n < 2:
+                    return 1
+                return fib(n-1) + fib(n-2)
+            fib(5)
 
-if get_tracer() is not None:
-    get_tracer().set_sync_marker()
+            if get_tracer() is not None:
+                get_tracer().set_sync_marker()
 
-"""
+        """)
 
         def expect_sync_marker(data):
             self.assertIsNotNone(data['viztracer_metadata'].get('sync_marker'))
@@ -381,18 +381,18 @@ if get_tracer() is not None:
 
     @package_matrix(["~orjson", "orjson"])
     def test_align_combine_sync_marker(self):
-        test_script = """
-from viztracer import get_tracer
+        test_script = textwrap.dedent("""
+            from viztracer import get_tracer
 
-if get_tracer() is not None:
-    get_tracer().set_sync_marker()
+            if get_tracer() is not None:
+                get_tracer().set_sync_marker()
 
-def fib(n):
-    if n < 2:
-        return 1
-    return fib(n-1) + fib(n-2)
-fib(5)
-"""
+            def fib(n):
+                if n < 2:
+                    return 1
+                return fib(n-1) + fib(n-2)
+            fib(5)
+        """)
 
         def expect_aligned_to_sync_marker(data, res1_filename, res2_filename):
             with open(res1_filename, 'r') as f:
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -411,13 +411,13 @@ class TestCommandLineBasic(CmdlineTmpl):
             self.assertEqual(len(func_2), 1)
             self.assertEqual(len(funcs), 2)
 
-            # we expect that unaligned events shifted more than 100ms
+            # we expect that unaligned events shifted more than 1s
             original_diff = abs(func_1[0]['ts'] - func_2[0]['ts'])
-            self.assertGreaterEqual(original_diff, 100000.0)
+            self.assertGreaterEqual(original_diff, 1000000.0)
 
-            # we expect that aligned events shifted not more than 0.1ms
+            # we expect that aligned events shifted not more than 50ms
             aligned_diff = abs(funcs[1]['ts'] - funcs[0]['ts'])
-            self.assertLessEqual(aligned_diff, 100.0)
+            self.assertLessEqual(aligned_diff, 50000.0, str(data))
 
         def test_align(extra_args):
             with tempfile.TemporaryDirectory() as tmpdir:
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -373,19 +373,19 @@ class TestCommandLineBasic(CmdlineTmpl):
             import time
             from viztracer import get_tracer
 
-            def func():
+            def test_func():
                 return 2 * 4
 
-            # will should sleep from 50ms to 100ms
+            # should sleep from 50ms to 100ms
             time.sleep(random.uniform(0.05, 0.1))
 
             get_tracer().set_sync_marker()
-            func()
+            test_func()
         """)
 
         def expect_aligned_to_sync_marker(data):
 
-            funcs = [event for event in data['traceEvents'] if 'ts' in event and event['name'].startswith('func ')]
+            funcs = [event for event in data['traceEvents'] if 'ts' in event and event['name'].startswith('test_func')]
             self.assertEqual(len(funcs), 2)
 
             # we expect that aligned events shifted not more than 1ms
--- a/tests/test_cmdline.py
+++ b/tests/test_cmdline.py
@@ -3,7 +3,9 @@
 
 import configparser
 import importlib.util
+import json
 import os
+import pathlib
 import re
 import signal
 import sys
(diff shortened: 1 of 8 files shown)


sergey-miryanov:
@gaogaotiantian should I do something with failed build? It seems not related to my changes.

gaogaotiantian:
No, it's failing all the CIs. I'll investigated that first.

gaogaotiantian:
This is related to https://github.com/tonistiigi/binfmt/issues/215

gaogaotiantian:
It's a docker image issue and fixed in #543. You can rebase or merge your branch so this will go away.

gaogaotiantian:
I took a quick look first. The overall approach is good, before I do a thorough review, just some direction suggestions.

I'm hoping that this feature touches less unnecessary code. You refactored some code - adding utilities and stuff, that might clean up the code a bit, but I think if it should be done, it should be done in a separate PR. Let's just do this sync marker feature as isolated as possible. Which means, probably just a `set_sync_marker` and a `get_sync_marker` method added for the C tracer. (And probably report a warning or something if the user tries to set sync marker twice, that normally means a mistake).

For alignment, I don't think we need an extra argument `--use_sync_marker`. I probably mentioned this, but no sync marker should be equivalent to set sync marker on the very first trace event - the result will not be off too much. So when we align reports, if we can find one, use that, otherwise do the original stuff, which is exactly equivalent to sync marker at first event. The user does not need to use an extra option to load sync marker - it should just work.

The testing is thorough, which is great, but maybe reduce the complexity a bit. First of all, we don't need to add extra tests for cases where sync markers are not used at all. Negative tests are sometimes helpful, but that's a lot of code to test when a feature is not enabled.

For the temporary files, I know there are different ways to deal with them in the code base, but most of them are artifacts from when I was new to the project :) I think the better way is to create a temporary directory with `tempfile.TemporaryDirectory` and use that as a context. Put everything there (except maybe the final `result.json` file which is overwhelmingly used among tests) and python will clean up the tempdir for you.

This is not a huge feature but not trivial either, don't feel rushed and let's land it properly :)

sergey-miryanov:
@gaogaotiantian Thanks for your review. Will make proposed changes.

sergey-miryanov:
@gaogaotiantian Please take a look. Tests still a bit complicated, maybe you can suggest some simplifications.

sergey-miryanov:
I believe I have couple ideas.

tests/test_cmdline.py
```diff
@@ -347,6 +349,108 @@ def test_combine(self):
...
+
+        self.template(
+            [sys.executable, "-m", "viztracer", "--ignore_frozen", "-o", "result.json", "cmdline_test.py"],
+            expected_output_file="result.json",
+            expected_stderr="Synchronization marker already set",
+            script=test_script,
+            check_func=expect_sync_marker,
+        )
+
+    @package_matrix(["~orjson", "orjson"])
```
gaogaotiantian, line 382: `package_matrix` actually costs plenty of time and it should only be enabled when necessary (for meaningful coverage). I don't think this specific feature requires the coverage for with and without `orjson`. Maybe just remove this?

gaogaotiantian (commented):
I think for implementation part you did great - just a few comments. The test case could be simplified a bit :)

tests/test_cmdline.py
```diff
@@ -347,6 +349,108 @@ def test_combine(self):
...
+    get_tracer().set_sync_marker()
+
+def fib(n):
+    if n < 2:
+        return 1
+    return fib(n-1) + fib(n-2)
+fib(5)
+"""
+
+        def expect_aligned_to_sync_marker(data, res1_filename, res2_filename):
```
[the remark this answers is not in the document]
sergey-miryanov: We can't guarantee that the timestamp of the first event and the sync-marker won't be the same. So, I tried to check for this, and if so, I do assertAlmostEqual.

But I got your idea.

sergey-miryanov:
Thank you for your review, will dig further :)

tests/test_cmdline.py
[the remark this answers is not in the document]
gaogaotiantian: Right, so we can do something like
```python
time.sleep(random.random())
set_sync_marker()
f()
```

Then the first timestamp is definitely not sync marker (the ts resoultion should be at ns level)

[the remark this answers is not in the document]
sergey-miryanov: I believe it is better now.

tests/test_cmdline.py
```diff
@@ -347,6 +348,93 @@ def test_combine(self):
...
+
+            def func(a, b):
+                return a * b
+
+            # will should sleep from 50ms to 100ms
+            rnd = random.random()
+            time.sleep(0.05 + 0.05 * rnd)
+
+            get_tracer().set_sync_marker()
+            func(rnd, rnd)
```
gaogaotiantian, line 397: It's okay to send `rnd` in, but like I said, it could confuse people - is it intentional, or just a random choice? I think we can make this simple and just do an empty func.

tests/test_cmdline.py
```diff
@@ -347,6 +348,93 @@ def test_combine(self):
...
+            funcs = [event for event in data['traceEvents'] if 'ts' in event and event['name'].startswith('func ')]
+            self.assertEqual(len(funcs), 2)
+
+            # we expect that aligned events shifted not more than 1ms
+            aligned_diff = abs(funcs[1]['ts'] - funcs[0]['ts'])
+            self.assertLessEqual(aligned_diff, 1000.0, str(data))
+
+        def test_align(extra_args):
+            with tempfile.TemporaryDirectory() as tmpdir:
+                res1_filename = pathlib.Path(tmpdir, 'res1.json').as_posix()
```
gaogaotiantian, line 411: I think `pathlib` should work in this case, but all of the other code uses `os.path`. The `as_posix()` piece actually confused me (I'm not that familiar with pathlib) and made me wonder - how would that work for Windows? But obviously it worked in CI. If there's no obvious advantage, could we use `os.path.join(tmpdir, "res2.json")` here?

For the command line below, I don't think you need the `--output_dir` (even though it's totally fine this way). I think you can simply do `"-o", res1_filename` to specify the absolute path.

tests/test_cmdline.py
sergey-miryanov: I think that sometimes, Python can optimize out calls of empty or trivial functions. But the more I think about this, the more I think that my version will be removed too. :)
So, you're right to keep it simple as we can.

tests/test_cmdline.py
sergey-miryanov: as_posix works well on all modern versions of Windows.

tests/test_cmdline.py
gaogaotiantian: I don't think Python ever does that - it would be a bug.

tests/test_cmdline.py
gaogaotiantian: Yeah it passed the CI so I think it should work. Just want to keep the consistency so we don't have two set of tools :)

sergey-miryanov: Landed :)

tests/test_cmdline.py
sergey-miryanov: Why not? Constant folding can fold trivial expressions into one constant. And inliner inline it :) Maybe some time :)

gaogaotiantian: That's true, but that would be a backward imcompatible change. I don't think CPython currently has a plan to skip/fold a function call if it's empty or trivial. Unlike C/C++ which has a language standard and the compiler can do whatever they want if it's not specified, Python is a de facto standard language, which means whatever is in it for a long enough time, becomes the standard and should not be changed. I believe even JIT is on frame-level now, which means it won't break the promise of a function call on the source code will always generates a function call.

sergey-miryanov: I saw a topic on discuss - https://discuss.python.org/t/inline-python-functions-and-methods/12412/10. Maybe I'm wrong and this is not related to us.

sergey-miryanov:
@gaogaotiantian build 3.12 on ubuntu-last failed with `Failed to CreateArtifact: Failed to make request after 5 attempts: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact`. Should I do something with this?

tests/test_cmdline.py
gaogaotiantian: Yeah that's from the implmentation perspective. Before 3.11, every Python call always corresponds to a C call, which takes some stack space and overheads for C functions. Since 3.11, a Python call can be "inlined" at C level, which means it does not actually generate a C call anymore, which is the major reason why 3.11 is much faster. However, from Python's perspective, everything is the same. You can't tell from Python level that the call did not actually happen at C level - all the triggers for a function call happened, a new frame is created, and you will be in the function (say you can step into it from a debugger).

So, yes it is kind of inlined, but that's an implementation detail. From Python's perspective, the call still happened.

gaogaotiantian:
> @gaogaotiantian build 3.12 on ubuntu-last failed with `Failed to CreateArtifact: Failed to make request after 5 attempts: Request timeout: /twirp/github.actions.results.api.v1.ArtifactService/CreateArtifact`. Should I do something with this?

I don't believe so, I'll just restart the CI.

tests/test_cmdline.py
```diff
@@ -347,6 +347,80 @@ def test_combine(self):
...
+                    common_cmd_line + ["-o", res2_filename, "cmdline_test.py"],
+                    expected_output_file=res2_filename,
+                    script=test_script,
+                    cleanup=False,
+                )
+
+                self.template(
+                    [sys.executable, "-m", "viztracer", "--align_combine", res1_filename, res2_filename],
+                    expected_output_file="result.json",
+                    script='',
```
gaogaotiantian, line 417: I don't think you need this `script` anymore.

tests/test_cmdline.py
```diff
@@ -347,6 +347,80 @@ def test_combine(self):
...
+        def expect_aligned_to_sync_marker(data):
+
+            funcs = [event for event in data['traceEvents'] if 'ts' in event and event['name'].startswith('func ')]
+            self.assertEqual(len(funcs), 2)
+
+            # we expect that aligned events shifted not more than 1ms
+            aligned_diff = abs(funcs[1]['ts'] - funcs[0]['ts'])
+            self.assertLessEqual(aligned_diff, 1000.0, str(data))
+
+        def test_align(extra_args):
```
gaogaotiantian, line 395: Last comment and we are ready to go.

I think we can simply do the for loop here. We don't need to create a function that takes a very trivial argument. Just do `for extra_args in [[], ["--dump_raw"]]:` and the following stuff.

sergey-miryanov: Ok, I will do this tomorrow- it's a bit late here now :)

tests/test_cmdline.py
sergey-miryanov: Thanks for explanation. I appreciate it!

tests/test_cmdline.py
gaogaotiantian: No worries, have a great night!

sergey-miryanov: Thanks!

tests/test_cmdline.py
```diff
@@ -347,6 +347,76 @@ def test_combine(self):
...
+    def test_align_combine_sync_marker(self):
+        test_script = textwrap.dedent("""
+            import random
+            import time
+            from viztracer import get_tracer
+
+            def func():
+                return 2 * 4
+
+            # will should sleep from 50ms to 100ms
```
gaogaotiantian, line 379: The comment should be fixed :)

tests/test_cmdline.py
```diff
@@ -347,6 +347,76 @@ def test_combine(self):
...
+            time.sleep(random.uniform(0.05, 0.1))
+
+            get_tracer().set_sync_marker()
+            func()
+        """)
+
+        def expect_aligned_to_sync_marker(data):
+
+            funcs = [event for event in data['traceEvents'] if 'ts' in event and event['name'].startswith('func ')]
```
gaogaotiantian, line 388: Is the reason we use `'func '` here (instead of `func`) that there are other events? Or just to make sure it's `func` function instead of a function starting with func? Random spaces in code sometimes will throw people off. I think the events should be pretty controllable right? Especially you ignored frozen. Can we remove the space or rename the function to something more unique?

gaogaotiantian:
Ahh and I forgot one thing - could you add some basic description in "combine reports" section of `concurrency.rst`? Just the usage of `set_sync_marker()`. Thanks!

sergey-miryanov:
> Ahh and I forgot one thing - could you add some basic description in "combine reports" section of `concurrency.rst`? Just the usage of `set_sync_marker()`. Thanks!

This is definitely the hardest part! 

I took part of the basic_usage.rst and added lines about sync-marker. I'm not a native speaker, so I'm not sure I can properly describe the desired actions in the docs.

gaogaotiantian:
Sorry that was my fault - it should not be in `concurrency.rst`. Just put the sync marker description after `align_combine` in basic usage. I forgot the existence of that piece :) Your description works - I'm not a native speaker either.

sergey-miryanov:
Please take a look.

gaogaotiantian:
Thank you for all the efforts! Will merge this soon.

sergey-miryanov:
Thanks for your patience!

PR #551, opened 2025-01-29
Backward compatible with even older PyTorch

yodahuang:
`traceName` was added in https://github.com/pytorch/kineto/pull/458. `viztracer` does not need the field, so it's safe skip dropping the field if the field was just not there. I verified that it works with PyTorch 1.11.


gaogaotiantian:
2021 is like 4 years ago and we don't really care about that anymore.

yodahuang:
Understood. I can patch the lib locally for my usecase.

PR #563, opened 2025-03-07
fix(decorator): fix trace and save in daemon process

HernandoR:
```
  File "/home/lz/.local/share/uv/python/cpython-3.10.16-linux-x86_64-gnu/lib/python3.10/multiprocessing/process.py", line 118, in start
    assert not _current_process._config.get('daemon'), \
AssertionError: daemonic processes are not allowed to have children
```
add condition if the current process is daemon

not sure why the multiprocessing.process uses `_config.get()` since to my testing `process.daemon` works fine. Maybe due to partly initiated object in some cases?


gaogaotiantian:
It's probably a correct fix but for viztracer I'd like a minimal reproducible example which could be turned into a regression test. If you want to work on the test that's perfect. However, if this is too much work for you, I'll see if I can repro this and fix it myself :)

PR #564, opened 2025-03-08
fix(decorator): fixed save in daemon mode

HernandoR:
Daemon processes are not allowed to have children so we need to save the trace file directly

follow up of #563 

added unittest, plz edit the test if it didn't play well with others :)


gaogaotiantian:
You don't need to close the other PR then create a new one. You can push to the same branch. Do you want me to just fix the test? I'll list you as the co-author. Your test has multiple issues - it will break on Windows, it probably will mess up with all the other multiprocessing tests and it should not be in its own file :)

HernandoR:
> You don't need to close the other PR then create a new one. You can push to the same branch. 

Yeah...the original PR was made on the github web, I used a local environment to develop the test, then noticed the 'patch-1' branch name. The PR was closed by github automatically when I renamed the branch to a meaningful name. I didn't realize that it would happen, sorry for any inconvenience.

> Do you want me to just fix the test? I'll list you as the co-author. Your test has multiple issues - it will break on Windows, it probably will mess up with all the other multiprocessing tests and it should not be in its own file :)

Yes please, I don't have a Windows machine for developing. and TBH, I didn't quite find the right place to insert this 
test. Anyway, this is an MVP to reproduce the issue and show the effect of the fix. Please integrate it with the tests' family.

PR #576, opened 2025-03-19
remove redundant code

qierenrongku:
The `new_children = []` variable does not seem to be used，And these `()` are redundant


gaogaotiantian:
There's nothing wrong with this PR, even though it's mostly cosmetic. However, this file is not used anymore. It's only kept to maintain backwards compatibility. I'd prefer keeping this file untouched. Thank you very much for the PR but I'll close it for now.

PR #589, opened 2025-05-11
Do not double patch multiprocessing + fork

gaogaotiantian:


gaogaotiantian:
This is not necessary because the finalizer will be cleared by _bootstrap.

PR #597, opened 2025-05-15
Call _threading_atexits in the reversed order

Chang-LeHung:
It is better to call functions in `_threading_atexits` in the reverse order. Cleanup functions are always "first in, last out". The `threading` module does this too ( in the function `_shutdown`, which is invoked by CPython internals when exiting).

### DIFF
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -531,5 +531,5 @@ class TestThreadingExitOrder(CmdlineTmpl):
         self.template(
             ["viztracer", "cmdline_test.py"],
             script=threading_exit_order,
-            expected_stdout=r"hello world"
+            expected_stdout="hello world"
         )
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -524,7 +524,7 @@ import threading
 
 
 def print_data(arg):
-    print(arg)
+    print(arg, end="")
 
 
 if __name__ == "__main__":
@@ -538,5 +538,5 @@ class TestThreadingExitOrder(CmdlineTmpl):
         self.template(
             ["viztracer", "cmdline_test.py"],
             script=threading_exit_order,
-            expected_stdout=r"1\n2\n"
+            expected_stdout=r"12"
         )
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -525,7 +525,7 @@ class TestThreadingExitOrder(CmdlineTmpl):
             import threading
 
             if __name__ == "__main__":
-                threading._register_atexit(print, " world", end="")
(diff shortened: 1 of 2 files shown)


gaogaotiantian:
Sure this makes sense, but you need a regression test for it. I would suggest using the test from issue141 as a template.

gaogaotiantian:
> Would you mind if I add result.html, result.json, sub.py, and .vscode/* to .gitignore? Those files are a little bit annoying.

You can do it in a seperate PR, but only `result.json` should be added. `result.html` is kind of deprecated and `sub.py` should be cleaned properly. Also, I never use `git add .` - I think that's in general a bad habit.

Chang-LeHung:
Hmm, I'd like to keep my git status clean. If that's not possible, then let's leave it as-is. 
This PR is ready for review. Please take a look when you're available.

Chang-LeHung:
I've tweaked the test case following your advice.

tests/test_regression.py
```diff
@@ -516,4 +516,20 @@ def task():
...
+                      expected_stdout="success")
+
+
+class TestThreadingExitOrder(CmdlineTmpl):
+    def test_threading_exit_order(self):
+        threading_exit_order = """
+            import threading
+
+            if __name__ == "__main__":
+                threading._register_atexit(print, " world", end="")
```
gaogaotiantian, line 528: I guess you don't need the `end=""` here.

tests/test_regression.py
```diff
@@ -516,4 +516,20 @@ def task():
...
+                threading._register_atexit(print, "hello", end="")
+            """
+        self.template(
+            ["viztracer", "cmdline_test.py"],
+            script=threading_exit_order,
+            expected_stdout=r"hello world"
```
gaogaotiantian, line 534: You also don't need the `r` here - it's just a very simple string.

Chang-LeHung: Yes

tests/test_regression.py
Chang-LeHung: If not, `\n` should be added to the pattern, but it does not work on Windows, and I do not have a machine to test it.

tests/test_regression.py
Chang-LeHung: I've removed it.

tests/test_regression.py
Chang-LeHung: Hmm, that makes sense.

Chang-LeHung: Done!

gaogaotiantian:
Thanks :)

PR #598, opened 2025-05-15
fix 591: addressing issues with forked processes

pulkin:
It sort of works with the example in the issue, except that (some of?) the parent process pre-fork stack is still shown for the forked process. I.e. it might take some mental capacity to understand when exactly the fork happened. Yet, I was able to finally profile a complex application with forking subprocesses using this patch.


gaogaotiantian:
Fixed in #601

Issue #599, opened 2025-05-18
The tid in the dumped metadata is incorrect


Chang-LeHung:
```diff
    // == Load the metadata first ==
    //    Process Name
    {
        PyObject* process_name = NULL;
        if (self->process_name) {
            process_name = Py_NewRef(self->process_name);
        } else {
            PyObject* current_process_method = PyObject_GetAttrString(multiprocessing_module, "current_process");
            if (!current_process_method) {
                perror("Failed to access multiprocessing.current_process()");
                exit(-1);
            }
            PyObject* current_process = PyObject_CallObject(current_process_method, NULL);
            if (!current_process_method) {
                perror("Failed to access multiprocessing.current_process()");
                exit(-1);
            }
            process_name = PyObject_GetAttrString(current_process, "name");
            Py_DECREF(current_process_method);
            Py_DECREF(current_process);
        }

-        fprintf(fptr, "{\"ph\":\"M\",\"pid\":%lu,\"tid\":%lu,\"name\":\"process_name\",\"args\":{\"name\":\"",
-                pid, pid);
        fprint_escape(fptr, PyUnicode_AsUTF8(process_name));
        fprintf(fptr, "\"}},");
        Py_DECREF(process_name);
    }

```
using pid as tid ❌.
Check it here: https://github.com/gaogaotiantian/viztracer/blob/master/src/viztracer/modules/snaptrace.c#L1469-L1471

PR #600, opened 2025-05-18
Fix the tid in the dumped metadata

Chang-LeHung:
Related issue https://github.com/gaogaotiantian/viztracer/issues/599

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -185,11 +185,10 @@ get_thread_id() {
     }
     return tid;
 #elif defined(__FreeBSD__)
-    info->tid = pthread_getthreadid_np();
+    return pthread_getthreadid_np();
 #else
-    info->tid = syscall(SYS_gettid);
+    return syscall(SYS_gettid);
 #endif
-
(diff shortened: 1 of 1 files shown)


gaogaotiantian:
This is metadata where tid does not mean anything. Calling `get_thread_id` in an unknown thread also won't give you a more "correct" result.

Chang-LeHung:
You're right that it doesn't affect the webpage, but shouldn't we either fix it or leave a comment for clarity? In fact, we could just remove the "tid" field altogether.

gaogaotiantian:
Chrome trace format is not very clear about if we can omit `tid` field. The documentation is not the best. I would rather not touch it if it's working.

Chang-LeHung:
It's safer to leave this unchanged since the test cases depend on this tid. If we remove it, we'll need to update the test cases accordingly. However, I've discovered another bug about tid in multi-process programs. I'll open an issue in a few days, haha! Finally, Happy PyCon! If you see this, please close this PR.

gaogaotiantian:
Sure, raise a new issue about the other multiprocessing bug.

Issue #604, opened 2025-05-24
TLS should be updated when the syscall fork is invoked


Chang-LeHung:
After fork, the `thread_info` should be updated, because the tid is inherited from its parent process. So the following happened:

![Image](https://github.com/user-attachments/assets/1363867f-dcac-4c0a-9ae9-498bc4da10e8)

This happens if we use:

- `os.fork`.
- `multiprocessing` with the start method `fork`.

`forkserver` will not lead to this, because we only patch, but not trace.
`spwan` will not lead to this as well, because we use `exec`.

I've fixed it:

![Image](https://github.com/user-attachments/assets/fe2abdda-827a-4612-aed2-2e2db8d4a891)
I'll open a PR later.

Code to reproduce:
`fork `
```python
import os


def hello():
    print("Hello World", os.getpid())


def child():
    print("Child process")
    hello()


def parent():
    print("Parent process")
    hello()


print("Hello World")
pid = os.fork()
if pid == 0:
    child()
else:
    parent()
    os.waitpid(pid, 0)

```
`multiprocessing` with 'fork'
```python
import multiprocessing
import os


def hello():
    print("Hello World", os.getpid())


def child():
    print("Child process")
    hello()


def parent():
    print("Parent process")
    hello()


print("Hello World")


if __name__ == "__main__":
    import multiprocessing as mp

    mp.set_start_method("fork")
    p = multiprocessing.Process(target=child)
    p.start()
    p.join()
    parent()

```

PR #605, opened 2025-05-24
Fix tls when the syscall fork is invoked

Chang-LeHung:
Issue: https://github.com/gaogaotiantian/viztracer/issues/604

I'll add the corresponding test case(s) later.

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -175,6 +175,24 @@ clear_stack(struct FunctionNode** stack_top) {
 // =============================================================================
 // Thread info related functions
 // =============================================================================
+static unsigned long
+get_thread_id()
+{
+#if _WIN32  
+    return GetCurrentThreadId();
+#elif defined(__APPLE__)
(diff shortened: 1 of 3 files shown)


gaogaotiantian:
What is the actual impact for this issue besids "it is theortically wrong"? This fix iterates all of the existing nodes, which could be millions and that introduces a very significant overhead after forking. I want to know if the benefit trumps the problem.

I mean, when we fork the process, the existing trace is actually from the old thread id - so there's no correct answer here. If it has some real problem we can figure out a way to fix that, otherwise it's kind of benign.

gaogaotiantian:
That did not answer the question about what we gain. I don't think we should clear the trace after we fork, that's pretty wrong.

gaogaotiantian:
Like I said, the result is not exactly "correct" if we "fix" it - viztracer keeps data from pre-fork and those functions did execute when the thread id was the previous value. It might be less wrong, but I don't see it as a trust issue.

You made a good point that we did this for multiprocessing+fork, which did take a while. I'll take a look at the code soon and see if it should be merged.

PR #614, opened 2025-09-03
Fix issues found by mypy 1.17.1

akx:
I was working on some stuff locally and noticed the version of mypy that got installed from requirements-dev had some complaints.

This tiny PR fixes those.


src/viztracer/patch.py
```diff
@@ -51,7 +51,7 @@ def build_command(args: Sequence[str]) -> list[str] | None:
                         py_args.append(f"-{cm_py_args}")
                     mode = [f"-{cm}"]
                     # -m mod | -mmod
-                    mode.append(cm_arg or next(args_iter, None))
+                    mode.append(cm_arg or next(args_iter))
```
gaogaotiantian, line 54: This is not a cosmetic change - this changes the behavior when `args_iter` exhausted. Is there a reason behind it? Is `mypy` complaining about it?

akx: Yes – the type of `mode` is implicitly `list[str]`, so you can't plonk a None in there.

```
src/viztracer/patch.py:54: error: Argument 1 to "append" of "list" has incompatible type "str | Any | None"; expected "str"  [arg-type]
```

src/viztracer/code_monkey.py
```diff
@@ -245,7 +245,7 @@ def get_string_of_expr(self, node: Union[ast.expr, ast.slice]) -> str:
             if isinstance(node.value, str):
                 return f"'{node.value}'"
             else:
-                return f"{node.value}"
+                return f"{node.value!r}"
```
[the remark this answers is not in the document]
akx: It is:
```
src/viztracer/code_monkey.py:248: error: If x = b'abc' then f"{x}" or "{}".format(x) produces "b'abc'", not "abc". If this is desired behavior, use f"{x!r}" or "{!r}".format(x). Otherwise, decode the bytes  [str-bytes-safe]
```

But yeah, you're right, we could just repr everything and be done with it.

src/viztracer/patch.py
```diff
@@ -51,15 +51,16 @@ def build_command(args: Sequence[str]) -> list[str] | None:
                         py_args.append(f"-{cm_py_args}")
                     mode = [f"-{cm}"]
                     # -m mod | -mmod
-                    mode.append(cm_arg or next(args_iter, None))
+                    if cm_arg := (cm_arg or next(args_iter, None)):
```
gaogaotiantian, line 54: I still don't think this is equivalent. This is a complicated piece of code. When we do lint fix, we don't want to change the behavior, because it will make code review much more difficult. So here's a suggestion - make a separate PR for cibuildwheel, that is super helpful. Then make a PR just to fix the mypy lint, and keep all the original behavior (unless it's super trivial). For example, do `list[str | None]` for now. Then we can discuss how to make this work better, without blocking all the PRs that would fail due to the easy stuff. Sorry I did not reply earlier, I'm travelling.

akx: I pushed one more change to tighten the walrus expression up in the case `cm_arg` ends up being the empty string (which is something of a malformed cmdline anyway).

gaogaotiantian: Ahh, I see. So I read the code again and tried to understand it. We can set `mode` to `None` for "invalid mode". So let's do something like 

```python
mode = None
...
cm_arg = cm_arg or next(arg_iter, None)
if cm_arg is None:
    mode = None
```

Then we can check if `mode` is `None` - it would be `None | list[str]`.

akx: Sure! Pushed that change.

gaogaotiantian:
And you need a coverage test for that. There is a special test for all the weird sub process arguments. You need one that has an empty -m.

akx:
> There is a special test for all the weird sub process arguments. 

`tests/test_patch.py`? I'm not quite sure what's happening there... :)

> You need one that has an empty -m.

As said in a previous comment, that'd be an invalid invocation anyway...

```
~ $ python3 -m ""
/opt/homebrew/opt/python@3.13/bin/python3.13: No module named
```

gaogaotiantian:
Yes that's invalid. Invalid invocation could be a test too. We need to make sure viztracer works properly when a bad argument is passed.

gaogaotiantian:
Thank you! That was great work. Now the CI should work properly again. I am traveling the world now do I don’t have enough time to maintain anything. I’m grateful for your efforts!

akx:
@gaogaotiantian Thank you! And thank you for viztracer, it's been an indispensable tool lately 😄  Happy travels! 🗺️

PR #617, opened 2025-09-17
CI: fix cibuildwheel

akx:
As advised in https://github.com/gaogaotiantian/viztracer/pull/614#discussion_r2346863493, this PR just fixes the CI cibuildwheel configuration.

The CI `lint` results are expected to be red, because the mypy changes from #614 aren't in this PR.


gaogaotiantian:
I don't think the regression test for this is necessary but it would be a good thing to have to avoid this kind of issue in the future. To break the loop I'll merge this now (as it looks very safe). Let's make the build green again and polish it later. Thank you for the PR.

PR #618, opened 2025-09-19
Speed up CI

akx:
* don't install unnecessary dependencies for cibuildwheel
* run lints only on the modernest Python

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,16 +7,13 @@ on:
 
 jobs:
   lint:
-    strategy:
-      matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: '3.13'
       - name: Install dependency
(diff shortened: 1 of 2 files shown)


akx:
> Passing lint on latest version does not guarantee the lint on older versions.

I think running lints on the latest version is fine, since tests are run on all supported versions. If there is e.g. new syntax (say, modern annotations) introduced in code, tests will fail on those old versions. WDYT – which lints do you think would be problematic?

gaogaotiantian:
We should not use latest syntax in our code base because we need to support old versions. We can only use the annotations that are supported by all of versions we support.

gaogaotiantian:
Not entirely. We have some lines that are protected by type checking so it will only execute on type check. Also lint failure is much faster so it can block the time consuming test and give faster feedback on the PR.

gaogaotiantian:
A successful run of `lint` takes about 15s. Removing `flake8` gives you 1 or 2 seconds at most. That's basically nothing in reality. The accumulation time does not mean anything because github action time is free for public repos. The only thing we care about is how fast we can get the feedback.

If I remember correctly, there is a maximum number of actions you can run simutaneously, add one extra action (by spliting `flake8`) might actually block something (or itself can be blocked).

From a maintenance point of view, this change adds one more place to update when we update new versions.

Like I said - it's just not worth it. There's no real value coming from this, just some theoretical numbers.

PR #620, opened 2025-09-29
Validate the output file extension on the command line

d-b-w:
The other day I monitored a 2hour long task on viztracer - unfortunately I gave the output file a silly extension so the process ended with an error and no recorded trace.

It would be nice to catch that requirement when reading user arguments


gaogaotiantian:
Thank you for finding and fixing this issue. I made a small change to the PR but otherwise its good!

Issue #612, opened 2025-08-02
FileNotFoundError when using --open flag with viztracer on Linux


XiaoWan-zi:
Hello, Tian Gao, I'm a Chinese fan who discovered you on Bilibili. It's great to meet you and become one of your fans!
Recently, when I use `viztracer --open` on Linux, I encounter an error:
```text
viztracer/src/viztracer/main.py", line 376, in run_code
    tracer.label_file_to_write()
viztracer/src/viztracer/viztracer.py", line 355, in label_file_to_write
    with open(output_file, "w") as _:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpwwl5o7ec/result_2967304.json.viztmp'
```

This error occurs whether I'm tracing a simple ` print('hello world')`  or any other Python file. I tried commenting out a section of code related to ` options.subprocess_child` :
```python
# if options.subprocess_child:
    # If it's a subprocess, we need to store the FEE data to the
    # directory from the parent process.
    # It's not practical to cover this line as it requires coverage
    # instrumentation on subprocess.
    # output_file = self.ofile  # pragma: no cover
# else:
output_file = os.path.join(self.multiprocess_output_dir, "result.json")
```
After this change, the web UI can be generated normally, but this might cause other bugs. My abilities are limited, and I haven't fully understood the real bug point yet. Could you please help confirm whether this is indeed a bug and what the root cause is? I'd like to try to fix this bug and contribute to the project. Thank you!
Sorry for the interruption, and I wish you a pleasant journey!

XiaoWan-zi:
bug

PR #621, opened 2025-10-01
fix --open bug

XiaoWan-zi:
I’ve come up with a solution for issue #612 . When you have time, could you take a look and see if it’s feasible?


gaogaotiantian:
Close because of #640

Issue #340, opened 2023-05-19
can not trace when a file contains os._exit()


yihong0618:
viztracer mypy # this work

viztracer mypy some.py # this not work

Minimize the reproduction of this program
cat a.py
```py
import os

def a():
    return 2

def b():
    return 3

def c():
    return a() + b()

c()
os._exit(0)
```
viztracer a.py # this generate nothing

gaogaotiantian:
This is not a bug. os._exit() terminates the process brutally so no exit hook has a chance to execute. You’ll experience the same issue with any library that requires an exit hook using this code, for example, coverage.

yihong0618:
Make sense and really thanks for your great tool.

PR #641, opened 2025-11-02
Add `os._exit` to limitations.rst

Otto-AA:
I needed a bit to figure out why my forked processes did not show up, until I had a repro and found #340 . So I think it makes sense to add this to the Limitations section.


yihong0618:
that's cool it reminds another issue thanks

PR #646, opened 2025-11-29
add -V to show version

XiaoWan-zi:
Hello, I’ve added `-V` to show the viztracer version and made a few minor optimizations. Please review the code, thanks.


gaogaotiantian:
There is already too many options for viztracer. `--version` is not a frequently needed option, we don't need a short option for it. Just delete it.

PR #647, opened 2025-12-01
add option `log_callsite` to record function callsite (filepath:lineno)

zzqq2199:
Overview
- Callsite is critical when analyzing complex code paths. For example, when I find the function `FA` takes a lot of time, I want to find where `FA` is triggered. 
- I add an option `log_callsite` to record function callsite (`filepath:lineno`).
- Disabled by default; when enabled, writes callsite into event metadata and shows in the report.

image

Design & Implementation
- Python API: `VizTracer(log_callsite=True)`.

### DIFF
--- a/src/viztracer/modules/eventnode.c
+++ b/src/viztracer/modules/eventnode.c
@@ -11,10 +11,11 @@ void
 clear_node(struct EventNode* node) {
     switch (node->ntype) {
     case FEE_NODE:
+        // Clear common optional fields
+        Py_CLEAR(node->data.fee.args);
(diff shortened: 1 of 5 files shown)


gaogaotiantian:
I believe viztracer once had this feature but the overhead was not ideal so I removed it. I still think this could be useful, but the implementation here is taking too much time. You are assembling strings in the trace callback, which is the most performance sensitive function in viztracer. (even if it's under a flag)

There should be no string manipulation during run time. All you need to do is store the integer. I don't think you need super fancy frame information because it will be obvious in Perfetto.

PR #676, opened 2026-03-31
Use zstd for multiprocess IPC compression when available

akx:
A lot of time was being spent in `zlib_compress`; on modern Pythons, we can use zstd for free.

Refs #674 (and conflicts with it).

### DIFF
--- a/src/viztracer/ipc_compression.py
+++ b/src/viztracer/ipc_compression.py
@@ -1,0 +1,28 @@
+# Licensed under the Apache License: http://www.apache.org/licenses/LICENSE-2.0
+# For details: https://github.com/gaogaotiantian/viztracer/blob/master/NOTICE.txt
+
+import zlib
+
+_ZSTD_MAGIC = b"zstd"
+
+try:
+    # `compression.zstd` is available on Python 3.14+ when
+    # compiled in to the build.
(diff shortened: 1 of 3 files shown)


gaogaotiantian:
Do you have benchmark to support the idea?

akx:
@gaogaotiantian Sure. Looks to be 4% faster on a trivial small example, likely more on a real program:

```
$ hyperfine 'git checkout master && .venv/bin/viztracer foot.py' 'git checkout zstd-when-possible && .venv/bin/viztracer foot.py'
Benchmark 1: git checkout master && .venv/bin/viztracer foot.py
  Time (mean ± σ):      2.707 s ±  0.034 s    [User: 2.495 s, System: 0.201 s]
  Range (min … max):    2.664 s …  2.756 s    10 runs

Benchmark 2: git checkout zstd-when-possible && .venv/bin/viztracer foot.py
  Time (mean ± σ):      2.603 s ±  0.022 s    [User: 2.393 s, System: 0.201 s]
  Range (min … max):    2.565 s …  2.643 s    10 runs

Summary
  git checkout zstd-when-possible && .venv/bin/viztracer foot.py ran
    1.04 ± 0.02 times faster than git checkout master && .venv/bin/viztracer foot.py
```

PR #677, opened 2026-04-26
Map CUPTI thread names to Python thread names to fix thread duplications

nh2:
Hi, I've been having the problem that with pytorch 2.3.0 with CUDA, `viztracer` `1.1.1` on Python 3.11.

When using CUDA via `pytorch` from 1 Python thread, all is great: I can see (in ui.perfetto.dev) the main thread calling `pytorch` functions, which in turn call torch `aten::` C++ functions, which in turn call `cudaLaunchKernel()` functions, all on the same thread track.

However, when I call CUDA from multiple Python threads (e.g. via `ThreadPoolExecutor`), it goes wrong:

1. The `aten::` functions disappear completely.
    * I am quite sure this is fixed by pytorch 2.4.0 (https://github.com/pytorch/kineto/issues/853, https://github.com/pytorch/pytorch/pull/123650). The issue was loss of precision because Kineto traces used microsecond instead of nanosecond granularity, and the conversion (rounding) caused chat child spans might no longer be contained in their parent spans, making ui.perfetto.dev reject them with parse errors.
2. The `cudaLaunchKernel` etc functions are not shown on the Python thread track they should be on, instead they appear in on a completely separate track (suggesting a separate thread), being disconnected from their Python/C++ callers. This makes reading traces difficult.

For (2), I managed to hack together a workaround, presented in this PR.

Note this was created with the Claude Opus 4.6 LLM; it apparently solves the problem, I read the code and it makes some sense to me, but do take as a caveat that I'm completely new to `viztracer`, `pytorch`, and CUPTI (though I do have general CUDA understanding).

Since I'm not 100% sure this is the correct solution, making this a draft PR.

The screenshot shows the problem with the extra threads:

image

And this is what it looks like fixed with my PR doing the remapping, with the extra threads gone:

image

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -4,6 +4,7 @@
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
 #include <stdlib.h>
+#include <string.h>
 #include <frameobject.h>
 #if _WIN32
 #include <windows.h>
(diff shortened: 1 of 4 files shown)


nh2:
Minimal repro I generated to test this across pytorch 2.3.0 and 2.6.0.

https://gist.github.com/nh2/1f61a60779dfa5fc303720dbb017dc81/7e6430d8b2be57ee6cfd455f6298d8961d0f0e10

PR #688, opened 2026-08-13
Don't call into CPython from the thread destructor after finalization

### LINKED ISSUE #687: snaptrace_threaddestructor calls PyGILState_Ensure() after interpreter finalization, causing SIGSEGV
# snaptrace_threaddestructor calls PyGILState_Ensure() after interpreter finalization, causing SIGSEGV

## Summary

`snaptrace_threaddestructor` is registered as a pthread TLS destructor
(`snaptrace.c:2045`). It is therefore invoked by glibc whenever a traced thread
terminates — including when that happens *after* `Py_FinalizeEx()` has already
torn down the interpreter. The function calls `PyGILState_Ensure()`
unconditionally (`snaptrace.c:285`), which dereferences interpreter state that
no longer exists, and the process dies with SIGSEGV (exit 139).

`tracer.stop()` and `tracer.save()` both return successfully first — the trace
file is written correctly — so the crash looks disconnected from tracing and
only shows up as a non-zero exit code after all real work has finished.

## Environment

- Linux x86_64 (Debian trixie), CPython 3.12.13
- Reproduced on VizTracer 1.1.1 (latest on PyPI) and on current `master`
  (293a7e3) — `snaptrace_threaddestructor` is unchanged between the two
- Reproduced in a container; nothing exotic in the environment

## Reproduction

```python
import threading
import time

from viztracer import VizTracer


def busy():
    deadline = time.time() + 2.0
    while time.time() < deadline:
        sorted({str(i): [i] for i in range(200)})


tracer = VizTracer(verbose=0, ignore_c_function=False, ignore_frozen=False)
tracer.start()

for _ in range(4):
    threading.Thread(target=busy, daemon=True).start()

time.sleep(0.3)
tracer.stop()
tracer.save("/tmp/repro.json")
print("stop() and save() returned; exiting with threads still running", flush=True)
```

```sh
for i in $(seq 1 30); do python repro.py >/dev/null 2>&1 || echo "crash"; done
```

**8 of 30 runs segfault on current `master`** (6 of 30 on 1.1.1), always after
the final `print()` has executed.
Joining the threads before exit eliminates it entirely (30/30 clean), which is
consistent with the destructor being safe as long as the thread dies while the
interpreter is still valid.

`ignore_c_function=True` does not help.

## Backtrace

Captured by attaching gdb after freezing the process at the fault (a raw
`SIGSTOP` from a `SIGSEGV` handler — a live ptrace-attached debugger perturbs
the timing enough to hide the race):

```
#0  0x... in ?? ()  from libpython3.12.so.1.0     <- consistent with new_threadstate
#1  0x... in ?? ()  from libpython3.12.so.1.0     <- consistent with PyGILState_Ensure
#2  snaptrace_threaddestructor (key=0x...) at src/viztracer/modules/snaptrace.c:285
#3  snaptrace_threaddestructor (key=0x...) at src/viztracer/modules/snaptrace.c:281
#4  0x... in ?? ()  from libc.so.6                <- pthread TLS destructor sweep
#5  0x... in ?? ()  from libc.so.6
#6  0x... in ?? ()  from libc.so.6                <- thread exit
```

Frame #2 resolves to exactly the `PyGILState_Ensure()` line. Fault address is
`0x60` — a small offset off a null pointer, i.e. a field read on an
interpreter/thread-state struct that is already gone.

At the same moment the main thread is in interpreter/process teardown — in one
capture inside `_PyPathConfig_ClearGlobal` (late `Py_FinalizeEx`), in another
inside `__cxa_finalize` (i.e. already past finalization, running shared-library
destructors during `exit()`).

## The deterministic variant

The daemon-thread case above is a race, but there is a common real-world
pattern where it is **100% reproducible**: an application that embeds a C++
extension module holding a worker thread pool as a process-lifetime `static`,
where those threads call `PyGILState_Ensure()` to hand results back to Python.

Such a thread is not joined when the job finishes — it is joined by the C++
static destructor chain that runs inside `exit()`. Because `main()` calls
`Py_RunMain()` (which finalizes the interpreter) and only *then* calls `exit()`,
that thread is guaranteed to die after finalization. Once VizTracer has traced
it even once, the crash is not probabilistic.

We observed exactly this in a production Python 3.12 service: **6 out of 6 runs
crashed** with the tracer enabled, and 0 out of 6 with it disabled, everything
else identical. The main thread's stack at crash time was:

```
__libc_start_main → exit() → [static destructors] → ~ConcurrentHashMap<…>
  → ~ThreadPoolExecutor → stopAndJoinAllThreads → std::thread::join()
    → pthread_join   [blocked, waiting on the very thread that is crashing]
```

## Suggested fix

Skip the cleanup when the interpreter is finalizing. `snaptrace.c` already
includes `pythoncapi_compat.h`, which provides a portable `Py_IsFinalizing()`
for every supported version, so no version guard is needed:

```c
if (info) {
    if (Py_IsFinalizing()) {
        return;
    }
    PyGILState_STATE state = PyGILState_Ensure();
    ...
```

The body of the destructor only does `Py_CLEAR()` and `PyMem_FREE()` on
per-thread buffers, so the early return just leaks them moments before the
process exits — which is what the CPython C API leaves you able to do at that
point anyway.

This is the same guard that other native code commonly uses in the same
position, e.g. deleters that need the GIL to drop a `pybind11::object` check
`Py_IsFinalizing()` and discard rather than attempt cleanup.

I have this patched and verified locally and will open a PR:

| | stock 1.1.1 | patched |
|---|---|---|
| minimal repro above (master) | 8/30 crash | **0/30** |
| production service (1.1.1) | 6/6 crash | **0/6** |

Tracing still works normally with the patch (trace files written, ~566k events,
valid JSON), and the application's own output is byte-for-byte identical to an
untraced run.

Happy to adjust the approach if you'd prefer a different one.

jakezur2:
Fixes #687

## Problem

`snaptrace_threaddestructor` is registered as a pthread TLS destructor
(`snaptrace.c:2045`), so it runs whenever a traced thread exits — including
after `Py_FinalizeEx()` has already torn the interpreter down. It then calls
`PyGILState_Ensure()` unconditionally, dereferences interpreter state that no
longer exists, and the process dies with SIGSEGV.

`tracer.stop()`/`tracer.save()` complete normally first, so the trace is written
and the crash only surfaces as a non-zero exit status afterwards.

Two ways to hit it:

- **Racy:** daemon threads still running when the main thread exits. 8 of 30
  runs crash on current `master` with CPython 3.12.13 (6 of 30 on 1.1.1).
- **Deterministic:** a native (C++) thread pool held as a process-lifetime
  `static` that calls `PyGILState_Ensure()` to return results to Python. It is
  joined by the C++ static-destructor chain inside `exit()`, which by
  construction runs after `Py_RunMain()` has already finalized the interpreter,
  so such a thread *always* runs its TLS destructor too late. 6 of 6 runs crash
  in a production service of this shape.

## Fix

Return early if the interpreter is finalizing. `snaptrace.c` already includes
`pythoncapi_compat.h`, which supplies a portable `Py_IsFinalizing()` across all
supported versions, so no version conditional is needed.

The destructor body only does `Py_CLEAR()` / `PyMem_FREE()` on per-thread
buffers, so the early return leaks those buffers microseconds before the process
exits — the only thing that is actually safe to do at that point.

## Verification

| | unpatched | patched |
|---|---|---|
| Minimal daemon-thread repro (`master`) | 8/30 crash | **0/30** |
| Production service, C++ pool (1.1.1) | 6/6 crash | **0/6** |

Tracing is unaffected: trace files are still written (~72 MB, ~566k events,
valid JSON), and the traced application's own output artifact is byte-for-byte
identical (`sha256` match) to an untraced run.

Full suite on this branch: `243 passed, 7 skipped, 2 subtests passed`.
`ruff check` and `ruff format --check` are clean.

## Test

Adds `TestThreadDestructorAfterFinalize` to `tests/test_regression.py`, which
runs the daemon-thread script as a subprocess and asserts a clean exit. Because
whether a worker exits before or after `Py_FinalizeEx()` is a race, it repeats
10 times; with the bug present that gives a high chance of catching it, and with
the fix it is stable. Skipped on Windows, where the TLS destructor is not
registered (`TlsAlloc` branch).

### DIFF
--- a/src/viztracer/modules/snaptrace.c
+++ b/src/viztracer/modules/snaptrace.c
@@ -281,6 +281,15 @@ snaptrace_threaddestructor(void* key) {
     struct ThreadInfo* info = key;
     struct FunctionNode* tmp = NULL;
     if (info) {
+        // A thread can exit after the interpreter is gone -- for example a
+        // native thread pool that outlives Py_FinalizeEx() and is joined by a
+        // C++ static destructor during exit(). There is no interpreter left
+        // for PyGILState_Ensure() to attach to, so it would dereference
+        // torn-down state and segfault. Leak the per-thread buffers instead;
+        // the process is already on its way out.
+        if (Py_IsFinalizing()) {
+            return;
+        }
         PyGILState_STATE state = PyGILState_Ensure();
         info->paused = 0;
         info->curr_stack_depth = 0;
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -659,3 +659,43 @@ class TestThreadingExitOrder(CmdlineTmpl):
             script=threading_exit_order,
             expected_stdout="hello world",
         )
+
+
+class TestThreadDestructorAfterFinalize(CmdlineTmpl):
+    # snaptrace_threaddestructor runs from the TLS destructor when a traced
+    # thread exits, which can be after the interpreter is finalized. It used to
+    # call PyGILState_Ensure() unconditionally and segfault. Daemon threads
+    # still running when the main thread exits reproduce the race.
+    @unittest.skipIf(
+        sys.platform == "win32", "TLS destructor is only registered on POSIX"
+    )
+    def test_thread_exiting_after_finalize(self):
+        thread_destructor_code = """
+            import threading
+            import time
+
+            from viztracer import VizTracer
+
+            def busy():
+                deadline = time.time() + 2.0
+                while time.time() < deadline:
+                    sorted({str(i): [i] for i in range(200)})
+
+            tracer = VizTracer(verbose=0, ignore_c_function=False, ignore_frozen=False)
+            tracer.start()
+            for _ in range(4):
+                threading.Thread(target=busy, daemon=True).start()
+            time.sleep(0.3)
+            tracer.stop()
+            tracer.save()
+            print("saved")
+            """
+        # Whether a worker exits before or after Py_FinalizeEx() is a race, so
+        # repeat to make a reintroduced crash reasonably likely to be caught.
+        for _ in range(10):
+            self.template(
+                [sys.executable, "cmdline_test.py"],
+                script=thread_destructor_code,
+                expected_output_file="result.json",
+                expected_stdout="saved",
+            )


gaogaotiantian:
Very detailed investigation and simple fix. I'll merge it soon.

Issue #682, opened 2026-06-30
[BUG] Yocto on aarch64, LICENSE and attach_linux_amd64.so


wenuam:
Hello, could you :

* fix your LICENSE scheme (PEP 621 vs PEP 639)
* not bundle AMD64 files in the aarch64 wheel

Here's the recipe I did to get it on the target though :

[python3-viztracer_1.1.1.bb.txt](https://github.com/user-attachments/files/29517090/python3-viztracer_1.1.1.bb.txt)

Regards

wenuam:
Hello,

I wrote a full recipe to compile `viztracer` on `aarch64` :

In folder `python3-viztracer` : `python3-viztracer_1.1.1.bb`

[python3-viztracer_1.1.1.bb.txt](https://github.com/user-attachments/files/31433934/python3-viztracer_1.1.1.bb.txt)

In subfolder `files` : `fix_inline.h`

[fix_inline.h.txt](https://github.com/user-attachments/files/31434006/fix_inline.h.txt)

```
RDEPENDS:${PN} = "\
    python3-viztracer \
    gdb \
"
```

`gdb` is mandatory for `attach_aarch64.so` to attach to the process.

Regards.

PR #691, opened 2026-08-24
Fix attach binary selection for cross-built wheels

mameikagou:
## Summary

- select the Linux attach helper from the wheel target platform instead of the build host
- exclude x86-only attach helpers from non-x86 wheels, including wheels built from an sdist
- preserve direct-build behavior and add platform-selection regression tests

## Testing

- `python3.11 -m unittest -v tests.test_setup_helpers`
- `uvx --offline ruff check setup.py setup_helpers.py tests/test_setup_helpers.py`
- `uvx --offline ruff format --check setup.py setup_helpers.py tests/test_setup_helpers.py`
- built clean source and sdist wheels for `linux_aarch64` and `linux_x86_64`; aarch64 contains no Linux attach helper, while x86_64 contains only `attach_linux_amd64.so`
- verified a direct host build still includes the AMD64 helper

Addresses the binary packaging part of #682.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,6 +2,7 @@ include LICENSE
 include NOTICE.txt
 include README.md
 include setup.py
+include setup_helpers.py
 
 recursive-include src/viztracer/html *.css *.js
 recursive-include src/viztracer/modules *.c *.h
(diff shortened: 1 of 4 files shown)


gaogaotiantian:
I don't think this is needed. With `cibuildwheel` the existing `setup.py` `package_data` should work properly. Let me know if I'm wrong.

